PR #13, opened 2021-10-27
rpc/eth: add roundParamFromBlockNum

### LINKED ISSUE #9: Handling `blockNum` parameter
Various web3 methods accept  the `blockNum` paramter, which can be one of the following:
```
Number: A block number
"earliest" - String: The genesis block
"latest" - String: The latest block (current head of the blockchain)
"pending" - String: The currently mined block (including pending transactions)
```

Currently most methods use the upstream `ethrpc.BlockNum` to handle the conversion from `blockNum` to a number, and then use it directly in the methods querying the oasis node, e.g.:
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/825d855ab35ce813f744dab66cd78021fe1cb84c/rpc/eth/api.go#L145-L154

Looking at the `ethrpc.BlockNum` the following mapping from string to number is done:
https://github.com/ethereum/go-ethereum/blob/7c4cad064cf7a9f48f9d69c6a4f459ebe8bb19ab/rpc/types.go#L58-L64

```
	PendingBlockNumber  = BlockNumber(-2)
	LatestBlockNumber   = BlockNumber(-1)
	EarliestBlockNumber = BlockNumber(0)
```

However this doesn't correctly map to oasis round numbers meanings. The correct way to map this would be:
```
LatestBlockNunber = 0 // 0 represents the "latest" round in oasis-node
PendingBlockNumber = 0 // Same as latest, there's no way currently to query pending blocks/transactions
EarliestBlockNumber = 1 // Should be either 1, or the genesis round or even better, query the oasis-node for the earilest round available (TODO: not sure if oasis-node currently exposes this, need to check)
```

pro-wh:
fixes #9 

I've added a method `roundParamFromBlockNum` based on the description in #9 

- `LatestBlockNumber` -> `RoundLatest` -- issue said `0`, but I think that was mistaken
- `PendingBlockNumber` -> `RoundLatest` -- using the same as latest, as suggested in the issue
- `EarliestBlockNumber` -> `1` -- using the option with the simplest implementation for now

The issue said that 1 is the earliest, and I'm taking its word for it. I actually don't know what round "0" would mean on Oasis.

The function doesn't do any validation. Zero and other negative values are casted blindly to uint64.

Some parameters that were declared as `number string` are now changed to `blockNum ethrpc.BlockNumber`, matching the rest of the file.
As I understand the type from go-ethereum, the special values should still log in a friendly way.

### DIFF
--- a/rpc/eth/api.go
+++ b/rpc/eth/api.go
@@ -363,7 +363,9 @@ func (api *PublicAPI) GetBlockByHash(blockHash common.Hash, fullTx bool) (map[st
 	round, err := api.backend.QueryBlockRound(blockHash)
 	if err != nil {
 		api.Logger.Error("Matched block error, block hash: ", blockHash)
-		return nil, ErrInternalQuery
+		// Block doesn't exist, by web3 spec an empty response should be retuned, not an error.
+		// TODO: ensure that the error is actually that the block doesn't exist.
+		return nil, nil
 	}
 
 	blk, err := api.client.GetBlock(api.ctx, round)
@@ -385,7 +387,8 @@ func (api *PublicAPI) getRPCTransaction(dbTx *model.Transaction) (*utils.RPCTran
 	txIndex := hexutil.Uint64(dbTxRef.Index)
 	resTx, err := utils.NewRPCTransaction(dbTx, blockHash, dbTxRef.Round, txIndex)
 	if err != nil {
-		api.Logger.Error("Failed to NewRPCTransaction", "hash", dbTx.Hash)
+		api.Logger.Error("Failed to NewRPCTransaction", "hash", dbTx.Hash, "err", err)
+		return nil, ErrInternalQuery
 	}
 	return resTx, nil
 }
@@ -396,23 +399,31 @@ func (api *PublicAPI) GetTransactionByHash(hash common.Hash) (*utils.RPCTransact
 
 	dbTx, err := api.backend.QueryTransaction(hash)
 	if err != nil {
-		api.Logger.Error("Failed to QueryTransaction", "hash", hash.String())
-		return nil, ErrInternalQuery
+		api.Logger.Error("Failed to QueryTransaction", "hash", hash.String(), "err", err)
+		// Transaction doesn't exist, by web3 spec an empty response should be retuned, not an error.
+		// TODO: ensure that the error is actually that the transaction doesn't exist.
+		return nil, nil
 	}
 
 	if dbTx == nil {
-		err = errors.New("transaction not found with the hash: " + hash.Hex())
-		return nil, err
+		api.Logger.Error("transaction not found with the hash", "hash", hash.Hex())
+		// Transaction doesn't exist, by web3 spec an empty response should be retuned, not an error.
+		// TODO: ensure that the error is actually that the transaction doesn't exist.
+		return nil, nil
 	}
 	return api.getRPCTransaction(dbTx)
 }
 
 // GetTransactionByBlockHashAndIndex returns the transaction for the given block hash and index.
 func (api *PublicAPI) GetTransactionByBlockHashAndIndex(blockHash common.Hash, index uint32) (*utils.RPCTransaction, error) {
+	api.Logger.Debug("eth_getTransactionByBlockHashAndIndex", "hash", blockHash.Hex(), "index", index)
+
 	round, err := api.backend.QueryBlockRound(blockHash)
 	if err != nil {
 		api.Logger.Error("Matched block error, block hash: ", blockHash)
-		return nil, ErrInternalQuery
+		// Block doesn't exist, by web3 spec an empty response should be retuned, not an error.
+		// TODO: ensure that the error is actually that the block doesn't exist.
+		return nil, nil
 	}
 
 	blk, err := api.client.GetBlock(api.ctx, round)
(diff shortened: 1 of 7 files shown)


rpc/eth/api.go
```diff
@@ -441,7 +453,7 @@ func (api *PublicAPI) GetTransactionByBlockHashAndIndex(blockHash common.Hash, i
 // GetTransactionByBlockNumberAndIndex returns the transaction identified by number and index.
 func (api *PublicAPI) GetTransactionByBlockNumberAndIndex(blockNum ethrpc.BlockNumber, index uint32) (*utils.RPCTransaction, error) {
 	api.Logger.Debug("eth_getTransactionByBlockNumberAndIndex", "number", blockNum, "index", index)
-	blockHash, err := api.backend.QueryBlockHash(uint64(blockNum))
+	blockHash, err := api.backend.QueryBlockHash(api.roundParamFromBlockNum(blockNum))
```
pro-wh, line 457: I've changed these ones going to the backend indexer as well, because the parameter name is "round." Can someone familiar with that component confirm that passing a round number parameter this way is correct?

kostko:
So zero is actually a perfectly normal round (usually the genesis but not always). The earliest round should probably be the genesis round.

pro-wh:
ok I gotta learn how to get that

I made it a method of the api thing instead of a standalone function for this reason though :laughing:

kostko:
The earliest round could also take the pruning policy into account but I'm not sure if there is such a query. Maybe it should respect the Web3 gateway's pruning policy?

kostko:
> ok I gotta learn how to get that

For the genesis block there's the `GetGenesisBlock` in the client API.

pro-wh:
thanks, drafted in #14

anonymousGiga:
> The earliest round could also take the pruning policy into account but I'm not sure if there is such a query. Maybe it should respect the Web3 gateway's pruning policy?

Genesis round can be used as the earlist one because there is no pruning now. We will implement pruning later, and then we can modify the earliest number.

rpc/eth/api.go
ptrus: Looks like this ends up calling: https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/b9c9f564eaeee0d09998652d3838b72b87601283/storage/psql/psql.go#L125-L135

Which I think, currently won't handle the `RoundLatest` case correctly. Should probably add a check in that method if `round==RoundLatest` and handle it separately by returning the row with `max block_ref.round`.

Probably same for any other backend indexer methods.

ptrus: Fixed in https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/13/commits/eed7481d28aa7d0a79da558262c09f20ba515d95

ptrus:
I added support for latest round (https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/13/commits/eed7481d28aa7d0a79da558262c09f20ba515d95). Also did some minor cleanup in the psql tests.

And removed unused `QueryTransactionByRoundAndIndex` (as it doesn't handle the latest round case correctly ATM, but also isn't used by anything). Can also be added back later once it's actually used.

https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/13/commits/1cec638f9f957349902fcf9e035b3457c473dbcc - reverts some regression that were done in 6be9dbb where methods would now (incorrectly return errors instead of empty response in some cases)

rpc/eth/api.go
```diff
@@ -72,6 +67,21 @@ func NewPublicAPI(
...
+// roundParamFromBlockNum converts special BlockNumber values to the corresponding special round numbers.
+func (api *PublicAPI) roundParamFromBlockNum(blockNum ethrpc.BlockNumber) uint64 {
+	switch blockNum {
+	case ethrpc.PendingBlockNumber:
+		// Oasis does not expose a pending block. Use the latest.
+		return client.RoundLatest
+	case ethrpc.LatestBlockNumber:
+		return client.RoundLatest
+	case ethrpc.EarliestBlockNumber:
+		return 1
```
kostko, line 79: This should probably use the (hopefully cached) genesis block (and later take pruning into account). Could be done in a separate issue.

rpc/eth/api.go
```diff
@@ -506,7 +508,7 @@ func (api *PublicAPI) GetTransactionReceipt(txHash common.Hash) (map[string]inte
 		}
 	}
 	logs := logs2EthLogs(oasisLogs, txRef.Round, common.HexToHash(txRef.BlockHash), txHash, txRef.Index)
-	//blockNum:=new(big.Int).SetUint64(txRef.Round).String()
+	// blockNum:=new(big.Int).SetUint64(txRef.Round).String()
```
kostko: Remove?

storage/psql/psql.go
```diff
@@ -140,6 +120,17 @@ func (db *PostDb) GetBlockHash(round uint64) (string, error) {
 	return blk.Hash, nil
 }
 
+// GetLatestBlockHash queries block hash by latest round.
```
kostko:
```suggestion
// GetLatestBlockHash queries for the block hash of the latest round.
```

rpc/eth/api.go
ptrus: Yeah there is a separate PR opened for this by @pro-wh will include it there

kostko (approved):

rpc/eth/api.go
ptrus: This one https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/14 (although doesn't have caching implemented currently)

PR #16, opened 2021-10-27
add runtime-id flag

### LINKED ISSUE #11: RuntimeID should be configurable
Currently it's hardcoded:
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/1f9823ee48879663fa0c8843128d6f7bd0e36cd1/main.go#L29

pro-wh:
fixes #11 

added in roughly the same way the other flags are declared

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,10 +7,10 @@ import (
 	"os/signal"
 	"syscall"
 
-	"github.com/spf13/cobra"
 	"github.com/oasisprotocol/oasis-core/go/common"
 	cmnGrpc "github.com/oasisprotocol/oasis-core/go/common/grpc"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	"github.com/spf13/cobra"
 	"google.golang.org/grpc"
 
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -36,6 +34,8 @@ var (
 	nodeAddr string
 	// Ethereum network id
 	chainId uint
```
pro-wh, line 36: I only see one use each of `nodeAddr` and `chainId`, only to declare the var. they're not used in the actual system, and I don't see other APIs trying to retrieve the values either

synote: The chainId is not set into the srvConfig.

main.go
```diff
@@ -36,6 +34,8 @@ var (
...
+	// Runtime ID as used in oasis-net-runner..
```
kostko:
```suggestion
	// Runtime identifier.
```

main.go
ptrus: Fixed both `nodeAddr` & `chainId` are now used.

kostko (approved):

PR #17, opened 2021-10-27
rpc/eth: add GetStorageAt

### LINKED ISSUE #12: Implement the `eth_getStorageAt` method
The runtime should already support it via the `evm.Storage` query

pro-wh:
fixes #12, I think

implemented as suggested in that issue

### DIFF
--- a/rpc/eth/api.go
+++ b/rpc/eth/api.go
@@ -3,6 +3,7 @@ package eth
 import (
 	"context"
 	"errors"
+	"fmt"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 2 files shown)


rpc/eth/api.go
```diff
@@ -187,6 +187,11 @@ func (api *PublicAPI) GetBlockTransactionCountByNumber(blockNum ethrpc.BlockNumb
 	return &n
 }
 
+func (api *PublicAPI) GetStorageAt(address common.Address, position []byte, blockNum ethrpc.BlockNumber) ([]byte, error) {
```
pro-wh: anyone have some advice on the position parameter and return types? does our framework use some special type to make sure it serializes/deserializes right? these are meant to be 256-bit quantities, as I understand it

synote: the [specs](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/eth1.0-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=true&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false) uses string(32 bytes hex encoded) as the `position` param.
The position may need to name as

 ```
 GetStorageAt(address common.Address, position string, blockNum ethrpc.BlockNumber) (hexutil.Bytes, error)
 ```
 
 the types in hexutil can make sure types are serialized properly.
 
 I suppose in th evm the return bytes should always be a 32 bytes slice.

rpc/eth/api.go
```diff
@@ -169,6 +170,26 @@ func (api *PublicAPI) GetBlockTransactionCountByNumber(blockNum ethrpc.BlockNumb
...
+
+	ethmod := evm.NewV1(api.client)
+	// TODO: blockNum ignored as EVM wrapper doesn't support queries for past rounds:
+	// https://github.com/oasisprotocol/oasis-sdk/issues/590
+	res, err := ethmod.Storage(api.ctx, address[:], pos)
+	if err != nil {
+		api.Logger.Error("failed to query storage", "err", err)
+		return "", ErrInternalQuery
+	}
+	return hexutil.Encode(res), nil
```
ptrus: This now almost works in my e2e tests, only seems like web3 test expects leading zeroes to be removed:
```
AssertionError: expected '0x0000000000000000000000007b0ccf25763e8901e42245b7b902218386a61b7c' to equal '0x7b0ccf25763e8901e42245b7b902218386a61b7c'
```

These tests pass on other ETH networks, so we should probably do as the tests expect.

rpc/eth/api.go
pro-wh: thanks, I'll be experimenting with hexutil.Big to eliminate leading zeros

rpc/eth/api.go
pro-wh: I pushed an update to encode the output as a number, so that leading zeros will be excluded. can you test with the new version?

rpc/eth/api.go
ptrus: Thanks, this works in my tests now

ptrus (approved):

Issue #10, opened 2021-10-27
Missing method `eth_getLogs`


ptrus:
Seems like the `eth_getLogs` is not implemented.

pro-wh:
I'm taking a look at this one. not assigning self yet though

ptrus:
this was done

PR #18, opened 2021-10-27
rpc/eth: add GetLogs

pro-wh:
for #10 

This splits out most of getRPCBlock into a getRPCBlockData method, which gives the logs. There might be some wasted work to get the information other than the logs. Feel free to tighten up the splitting in later work.

I wrote up some logic to walk through the block boundaries. but it's not tested and it's coupled with some expensive query stuff, so it's going to be hard to verify.

In this preliminary implementation, we don't have all the filtering features, notably for addresses and topics.

In this implementation, it's an error to request any blocks that don't exist yet. You don't get the logs from existing blocks when that error occurs. I don't know if that's consistent with the spec. Going from an earlier block up to "latest" is intended to work though.

### DIFF
--- a/rpc/eth/api.go
+++ b/rpc/eth/api.go
@@ -6,10 +6,10 @@ import (
 	"fmt"
 	"math/big"
 
-	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/eth/filters"
 	"github.com/ethereum/go-ethereum/rlp"
 	ethrpc "github.com/ethereum/go-ethereum/rpc"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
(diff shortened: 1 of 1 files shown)


rpc/eth/api.go
```diff
@@ -586,6 +596,61 @@ func (api *PublicAPI) GetTransactionReceipt(txHash common.Hash) (map[string]inte
...
+		if filter.ToBlock != nil {
+			round, err := api.roundParamFromBlockNum(ethrpc.BlockNumber(filter.ToBlock.Int64()))
+			if err != nil {
+				return nil, fmt.Errorf("convert to block number to round: %w", err)
+			}
+			endRoundInclusive = round
+		}
+	}
+
+	// Warning: this is unboundedly expensive
```
ptrus, line 677: Hm yeah i guess ideally these would be indexed and queried from the postgress db used by the gateway

kostko: Yes this is should be served from the PostgreSQL index, not by querying the node.

ptrus: The needed data is not even indexed at the moment, will open an issue for it.

ptrus: I guess this can be merged, and handle the serving from index in the issue tracking that:https://github.com/starfishlabs/oasis-evm-web3-gateway/issues/19

ptrus (approved):

Issue #19, opened 2021-10-28
Index more things (blocks, logs, transaction receipt info)


ptrus:
Currently only transactions are indexed and for blocks only a round-hash mapping is stored:
```
// BlockRef represents the relationship between block round and block hash.
type BlockRef struct {
	Round uint64 `pg:",pk"`
	Hash  string
}
```
Instead, evm blocks should be indexed as well.

For example the `eth_getBlockByHash` method currently does the following:
- queries the `BlockRef` table to get the `Round` for the `Hash`
- uses the round to query the block from the oasis-node
- parses the oasis-node block into an ethereum block on the fly

Instead if the ethereum block was indexed as part of the indexer, the `eth_getBlockByHash` could do:
- query the `Blocks` table to get the ethereum block and return it

The full list of methods that should be updated:
- [x] `GetBlockByNumber` - should just query the block from the DB
- [x] `GetBlockTransactionCountByNumber` - should just query the transaction count from the DB
- [x] `GetBlockTransactionCountByHash` - should just query the transaction count from the DB
- [x] `GetBlockByHash` - should just query the block from the DB
- [x] `GetTransactionByBlockHashAndIndex` - should get block from the DB (also for some reason also queries oasis-node for transactions, which should already be indexed in the DB)
- [x] `GetTransactionReceipt` logs and other receipt info should be indexed, to avoid querying oasis-node and computing these on the fly
- [x] `BlockNumber` - should just get the latest block from the indexer

PR #21, opened 2021-10-28
add pruning for indexer

### LINKED ISSUE #2: Add removal of old indexed blocks to the indexer
There should be some way of (automatically) deleting old indexed blocks to prevent the database from filling up.

A config option should be added to specify the age limit of blocks.  Then periodic checks should be made to delete blocks older than the specified age limit.

There might be a better way to implement this (I can't remember if postgresql supports setting expiry time on rows or not), so feel free to use that if it's faster :)

### LINKED ISSUE #20: Methods shouldn't retrun an error when the requested block/tx doesn't exist
Methods such as ` eth_getTransactionByHash` / `eth_getTransactionReceipt` and others, should (as per [web3 spec](https://eth.wiki/json-rpc/API)) spec return an empy response and not an error in case the requested TX was not found.

For example specifically for `eth_getTransactionReceipt` it says:  
```
Returns:
Object - A transaction receipt object, or null when no receipt was found`
```

---

Concretely, during my testing, I encountered that if an `error` is returned instead of an empty response, the `truffle`/`web3`  stopped retrying the query for the `eth_getTransactionReceipt` as it was waiting for the transaction to be executed, which lead to invalid behavior.

Did a temporary fix for this in https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/13/commits/1cec638f9f957349902fcf9e035b3457c473dbcc

--- 

Should double-check how these methods are handled by Infura and similar and match those responses: i did check for `eth_getTransactionReceipt` & `eth_getBlockByNumber` and Infura returns no error:
```
{"jsonrpc":"2.0","id":1,"result":null}
```

anonymousGiga:
Add the pruning feature for indexer #2 , but the pruning function in our backend is not complete. We need to complete the issue #19  first and then  improve this function.

### DIFF
--- a/indexer/backend.go
+++ b/indexer/backend.go
@@ -65,6 +65,8 @@ type Backend interface {
 		txs []*types.UnverifiedTransaction,
 	) error
 
+	Purning(step uint64) error
+
 	Close()
 }
 
@@ -197,6 +199,16 @@ func (p *psqlBackend) Index(
 	}
 
 	p.logger.Info("Indexed block: ", "round", round)
+
+	return nil
+}
+func (p *psqlBackend) Purning(step uint64) error {
+	if err := p.storage.Delete(new(model.BlockRef), step); err != nil {
+		p.logger.Error("Purning from db error", "err", err)
+		return err
+	}
+
+	//TO DO: Will remove ethblock, logs, transactions latter
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


pro-wh (commented):
preliminary comments:

1. p**ur**ning -> p**ru**ning throughout
2. need detailed explanation of what "step" is. the explanation should be detailed enough to understand precisely what's deleted when, without having to look at the code

anonymousGiga:
As some words write error, i opened a new pr #26  and close this pr.

PR #25, opened 2021-10-28
rpc/eth: add block number TODOs for other evm queries

pro-wh:
add notes similar to the one in #17 about other evm module queries


ptrus:
This was addressed in https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/31 I think, feel free to reopen if any was missed

PR #26, opened 2021-10-29
Add pruning for indexer

anonymousGiga:
As some words write error, i closed #21 , and use this instead.

### DIFF
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -14,8 +14,9 @@ import (
 )
 
 const (
-	storageRequestTimeout = 5 * time.Second
-	storageRetryTimeout   = 1 * time.Second
+	storageRequestTimeout    = 5 * time.Second
+	storageRetryTimeout      = 1 * time.Second
+	CheckPruningTimeInterval = 60 * time.Second
 )
 
 const RoundLatest = client.RoundLatest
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -50,7 +60,9 @@ var logger = logging.GetLogger("evm-gateway")
 func init() {
 	rootCmd.Flags().StringVar(&nodeAddr, "addr", nodeDefaultAddress, "address of node socket path")
 	rootCmd.Flags().UintVar(&chainId, "chainid", 42261, "ethereum network id")
-	rootCmd.Flags().StringVar(&runtimeIDHex, "runtime-id", "8000000000000000000000000000000000000000000000000000000000000000", "runtime id in hex")
+	rootCmd.Flags().StringVar(&runtimeIDHex, "runtime-id", defaultRuntimeIdHex, "runtime id in hex")
+	rootCmd.Flags().BoolVar(&enablePruning, "enablePruning", false, "enable pruning feature for indexer")
```
ptrus: I moved all settings in the config in a recent PR: https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/65a0c69ca9a98bfb3029791f719a6522fed9c0d3/conf/config.go#L11-L17 

maybe move these settings there as well.

indexer/indexer.go
```diff
@@ -65,6 +68,24 @@ func (s *Service) getRoundLatest() (uint64, error) {
 	return blk.Header.Round, nil
 }
 
+func (s *Service) puring() {
+	for {
+		if s.stopFlag {
```
ptrus: This was already "wrong" before. This `stopFlag` should probably be an `atomic` bool as it's set in a different go-routine than in which it's checked, and a normal `bool` is not thread-safe.

However, even better would be to remove the `stopFlag` as it is not actually needed, as you can use `s.ctx` to detect when you need to exit, since that context will be canceled by the "parent" go-routine:

```golang
func (s *Service) puring() {
    for {
        select {
            case <-s.ctx.Done():
                // Context canceled, end.
	        return 
            case <-time.After(CheckPruningTimeInterval):
        }
   
        // Do the prunning...
    }
}
```

Same change should be applied to the `periodIndexWorker` routine.

anonymousGiga: Ok, thanks for your help.

SzeChuanChilliSauce (approved):

Issue #28, opened 2021-10-29
Server config path should be configurable


ptrus:
The server config path should be configurable with a `--config` flag, and not harcoded to `conf/server.yml`: 
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/6c85c5b2346bd8853867e9365505ed762d31cad6/main.go#L86-L87

---

Additionally, let's move all other flags to the config as well, so that a yaml file can be used to configure all of the settings:
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/6c85c5b2346bd8853867e9365505ed762d31cad6/main.go#L50-L54

PR #29, opened 2021-10-29
enable configuring everything via config.yml file

### LINKED ISSUE #27: Support setting server configuration
Gateway server configuration is currently hardcoded to the default configuration, with no way to update by the user: https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/e6c8bf031643898d629f0fc5cac68bca709bc93f/main.go#L110

Should probably move server configuration into `conf/config.go`: https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/e6c8bf031643898d629f0fc5cac68bca709bc93f/conf/config.go#L9-L11

so that server configuration can be set in the yaml file.

ptrus:
Fixes #27 #28 

Makes everything configurable via the config.yml file.

Should maybe use some configuration framework in future, which would allow a combination of yaml and flags for configuration, but this should work for now.


kostko (approved):
Nice! In the future this should be refactored to use `viper` (since it is already used by Oasis Core et al.) instead of the yaml package.

PR #24, opened 2021-10-28
rpc/eth: support eth_estimateGas with updated version of oasis-sdk

synote:
Support gas estimation with oasis `EstimateGasForCaller`

Maybe I mistaken the fee structure of oasis Transaction

PR #31, opened 2021-10-29
rpc/eth: support eth_estimateGas with updated version of oasis-sdk

ptrus:
Taken from https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/24 + some (unrelated fixes)


synote (approved):
I think the estimateGas now can work with web3js `buildTransaction` in contract

PR #33, opened 2021-10-30
indexer: Take last retained round into account

### LINKED ISSUE #32: Query last retained block to determine start round
The Oasis Core node that the Web3 gateway is connecting to may not have all historic blocks available (e.g. due to using state sync) so once https://github.com/oasisprotocol/oasis-core/pull/4334 lands, is backported to 21.3.x and supported in the Oasis SDK, the Web3 gateway should use the `GetLastRetainedBlock` method to determine which round to start syncing from as syncing any earlier blocks will fail.

kostko:
Fixes #32 

Also adds some basic CI, although the tests are currently not even building.


indexer/backend.go
```diff
@@ -65,7 +65,10 @@ type Backend interface {
 		txs []*types.UnverifiedTransaction,
 	) error
 
-	Pruning(step uint64) error
+	Prune(round uint64) error
```
ptrus, line 70: Maybe also add a doc comment

ptrus (approved):

PR #36, opened 2021-11-02
api/SendRawTransaction: propagate CheckTx errors

ptrus:
Note: [oasis-core 23.1.4](https://github.com/oasisprotocol/oasis-core/blob/v21.3.4/CHANGELOG.md#2134-2021-10-30) is needed so that the client node propagates the errors.

### DIFF
--- a/rpc/eth/api.go
+++ b/rpc/eth/api.go
@@ -39,6 +39,7 @@ var (
 	ErrBlockNotFound              = errors.New("block not found")
 	ErrTransactionNotFound        = errors.New("transaction not found")
 	ErrTransactionReceiptNotFound = errors.New("transaction receipt not found")
+	ErrMalformedTransaction       = errors.New("malformed transaction")
 )
 
 // Log is the Oasis Log.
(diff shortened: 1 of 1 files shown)


rpc/eth/api.go
```diff
@@ -371,6 +371,13 @@ func (api *PublicAPI) Call(args utils.TransactionArgs, blockNum ethrpc.BlockNumb
 func (api *PublicAPI) SendRawTransaction(data hexutil.Bytes) (common.Hash, error) {
 	api.Logger.Debug("eth_sendRawTransaction", "length", len(data))
 
+	// Decode the Ethereum transaction.
+	ethTx := &ethtypes.Transaction{}
+	if err := rlp.DecodeBytes(data, ethTx); err != nil {
+		api.Logger.Error("Failed to decode rawtx data", "error", err.Error())
```
kostko:
```suggestion
		api.Logger.Error("failed to decode raw transaction data", "err", err.Error())
```

rpc/eth/api.go
```diff
@@ -382,13 +389,7 @@ func (api *PublicAPI) SendRawTransaction(data hexutil.Bytes) (common.Hash, error
 	err := api.client.SubmitTxNoWait(api.ctx, &utx)
 	if err != nil {
 		api.Logger.Error("Failed to SubmitTx", "error", err.Error())
```
kostko:
```suggestion
		api.Logger.Warn("failed to submit transaction", "err", err.Error())
```

kostko (approved):

PR #40, opened 2021-11-03
rpc: implement method web3_clientVersion

### LINKED ISSUE #39: Implement rpc method `web3_clientVersion`
the web api `web3_clientVersion` can return the build version info

synote:
`web3_clientVersion` could be used in web3js

### DIFF
--- a/rpc/web3/api.go
+++ b/rpc/web3/api.go
@@ -1,8 +1,12 @@
 package web3
 
 import (
+	"fmt"
+	"runtime"
+
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/starfishlabs/oasis-evm-web3-gateway/version"
 )
 
 // PublicAPI is the web3_ prefixed set of APIs in the Web3 JSON-RPC spec.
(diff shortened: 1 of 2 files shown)


rpc/web3/api.go
```diff
@@ -17,3 +20,9 @@ func NewPublicAPI() *PublicAPI {
 func (a *PublicAPI) Sha3(input hexutil.Bytes) hexutil.Bytes {
 	return crypto.Keccak256(input)
 }
+
+// ClientVersion returns the current client info.
+func (a *PublicAPI) ClientVersion() string {
+	goVersion := runtime.Version()
+	return fmt.Sprintf("oasis/%s", goVersion)
```
ptrus: Can probably avoid the separate variable and just do:
```golang
return fmt.Sprintf("oasis/%s", runtime.Version())
```

ptrus (approved):

rpc/web3/api.go
kostko: Why is this using the Go runtime version? Shouldn't this be something like the Web3 gateway version?

synote: Yeah, any advices for the gateway version, Use git revision or define a Version number?

kostko: I would define a version number (e.g. in a separate `version` package), let's set it to `0.1.0-alpha` for now. Later we can make this follow the release process.

synote:
The return of `web3_clientVersion` now is similar to https://web3js.readthedocs.io/en/v1.2.11/web3-eth.html#getnodeinfo

kostko (approved):

PR #44, opened 2021-11-04
Store more things in indexer

### LINKED ISSUE #1: Postgresql password should be moved into an environment variable
Currently, the postgresql password is stored in plaintext in the config file (`conf/server.yml`).

It should be moved into an environment variable (and the username into its own environment variable as well).

### LINKED ISSUE #10: Missing method `eth_getLogs`

### LINKED ISSUE #19: Index more things (blocks, logs, transaction receipt info)

### LINKED ISSUE #38: db Insert error return ignored, not handled
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/f6a1b3ce1622603f9f2226da63a6644a2891802c/indexer/backend.go#L187

ignores the potential error return from `p.storage.Store(blockRef)` -- this is possible, since

https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/f6a1b3ce1622603f9f2226da63a6644a2891802c/storage/psql/psql.go#L73-L76
passes up the error return from `db.Db.Model(value).Insert()`.

instead, `Index` only ever returns `nil`.

the user of this is https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/f6a1b3ce1622603f9f2226da63a6644a2891802c/indexer/indexer.go#L56

which would have returned `ErrIndexedFailed` if it ever got a non-`nil` status.  `indexBlock` is used from `indexingWorker()`, and errors cause the worker thread to sleep.  `indexBlock` returns two types of errors, the first from `s.client.GetBlock` or `s.client.GetTransaction` which are networking (grpc) issues that may be temporary.  the database error that _isn't_ currently returned is local and not a network problem, so it's unclear that sleeping is the right thing to do if/when the error propagation is plumbed through -- go-pg/pg seem to indicate that `Insert()` failures are panic-worthy, so maybe that's the right thing to do and require a wipe/rebuild of the db or restore from backup or something like that?  If a db corruption causes the `Insert()` to consistently fail, then the `nil` comparison in `indexingWorker` will live-lock, essentially just continuously loop sleeping and generating a misleading Warn log entry.

### LINKED ISSUE #59: nit: context.TODO() in initialization function
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/22eb39c828e9319dbbb9e14f9fe1be9192f8715f/storage/psql/psql.go#L33

Instead of `context.TODO()`, the `context` package recommends `context.Background()` for initialization, and `InitDb` is presumably an initialization function used only when the gateway is starting up.

If `InitDb` is going to be extended to accept a context so timeouts/cancellations can be done, then `context.TODO()` is fine, but there should be a `// TODO` comment explaining this at around L19.

anonymousGiga:
This is pr store ethblock, ethtransaction information in db. #19


SzeChuanChilliSauce (approved):

ptrus:
Thanks! I'll try to review this a bit later today.

lesterli (approved):

SzeChuanChilliSauce (approved):

storage/psql/psql.go
```diff
@@ -143,3 +143,112 @@ func (db *PostDb) GetContinuesIndexedRound() (uint64, error) {
...
+//		"from":              tx.FromAddr,
+//		"to":                tx.ToAddr,
+//	}
+//
+//	return receipt, nil
+//}
+
+func (db *PostDb) GetLogs(blockHash string) ([]*model.Log, error) {
+	logs := []*model.Log{}
+	err := db.Db.Model(&logs).Where("block_hash=?", blockHash).Select()
```
ptrus: Get logs should also support the following query: `BETWEEN startRound AND endRound`.

That way we can avoid looping through rounds in the `eth_getLogs` API method and get these with a single query.

rpc/eth/api.go
```diff
@@ -680,21 +587,13 @@ func (api *PublicAPI) GetLogs(filter filters.FilterCriteria) ([]*ethtypes.Log, e
...
-		block, err := api.client.GetBlock(api.ctx, round)
-		if err != nil {
-			if errors.Is(err, roothash.ErrNotFound) && round != client.RoundLatest && endRoundInclusive == client.RoundLatest {
-				// We've walked up to the latest round
-				break
-			}
-			return nil, fmt.Errorf("get block %d: %w", round, err)
-		}
-		_, _, _, blockLogs, err := api.getRPCBlockData(block)
+		dbLogs, err := api.backend.GetLogs(*filter.BlockHash)
```
ptrus: If the backend supports the `BETWEEN` query for the rounds (see the comment on the backend method) this loop can be replaced by a single query.

indexer/backend.go
```diff
@@ -55,15 +58,26 @@ type QueryableBackend interface {
 	Decode(utx *types.UnverifiedTransaction) (*model.Transaction, error)
 }
 
+type GetEthInfoBackend interface {
+	GetBlockByNumber(number uint64) (*model.Block, error)
+	GetBlockByHash(blockHash ethcommon.Hash) (*model.Block, error)
+	GetBlockTransactionCountByNumber(number uint64) (int, error)
+	GetBlockTransactionCountByHash(blockHash ethcommon.Hash) (int, error)
+	GetTransactionByBlockHashAndIndex(blockHash ethcommon.Hash, txIndex int) (*model.Transaction, error)
+	// GetTransactionReceipt(txHash ethcommon.Hash) (map[string]interface{}, error)
```
ptrus: Any reason why this one is not done, are there any issues for supporting this one from the DB?

model/init.go
```diff
@@ -8,6 +8,7 @@ import (
 // InitModel initializes db models.
 func InitModel(db *pg.DB) error {
 	models := []interface{}{
+		new(Block),
```
ptrus, line 11: Should `Log` be added here as well?

indexer/backend.go
SzeChuanChilliSauce: This is a temporary comment and we will fully implement it.

model/init.go
SzeChuanChilliSauce: yes, we will append the Log.

storage/psql/psql.go
SzeChuanChilliSauce: ok

rpc/eth/api.go
SzeChuanChilliSauce: ok

rpc/eth/api.go
```diff
@@ -17,15 +17,13 @@ import (
 	ethrpc "github.com/ethereum/go-ethereum/rpc"
```
ptrus, line 17: `getRPCBlockData` and `getgetRPCBlock` methods are not used anymore with this PR, so can be deleted from this file, right?

rpc/eth/api.go
```diff
@@ -17,15 +17,13 @@ import (
...
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
-	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
 	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
```
ptrus: The `ethrpc.EarliestBlockNumber` case in `roundParamFromBlockNum` should be updated to return the: `max(api.client.GetLastRetainedBlock(api.ctx), <database.LastRetainedBlock>)` 

So that the indexer pruning is taken into account (e.g. don't query pruned blocks)

rpc/utils/utils.go
```diff
@@ -174,3 +176,64 @@ func NewRPCTransaction(
 
 	return resTx, nil
 }
+
+func ConvertToEthBlock(block *model.Block) map[string]interface{} {
```
ptrus: Add a `includeFullTx bool` parameter, and in case it's `false` only include transaction hashes in the `transactions` field.

rpc/eth/api.go
```diff
@@ -501,20 +483,14 @@ func (api *PublicAPI) EstimateGas(args utils.TransactionArgs, blockNum *ethrpc.B
...
-
-	blk, err := api.client.GetBlock(api.ctx, round)
-	if err != nil {
-		api.Logger.Error("Matched block error, block round: ", round)
-		return nil, err
+		return nil, ErrInternalQuery
 	}
 
-	return api.getRPCBlock(blk)
+	return utils.ConvertToEthBlock(blk), nil
```
ptrus: Propagate the `fullTx bool` parameter here (also see the comment on the `utils.ConvertToEthBlock` method).

rpc/eth/api.go
```diff
@@ -180,33 +172,34 @@ func (api *PublicAPI) GetBlockByNumber(blockNum ethrpc.BlockNumber, _ bool) (map
...
+
+	blk, err := api.backend.GetBlockByNumber(round)
 	if err != nil {
 		api.Logger.Error("GetBlock failed", "number", blockNum, "error", err.Error())
 		// Block doesn't exist, by web3 spec an empty response should be returned, not an error.
 		return nil, ErrInternalQuery
 	}
 
-	return api.getRPCBlock(resBlock)
+	return utils.ConvertToEthBlock(blk), nil
```
ptrus: Propagate the `fullTx bool` parameter here (also see the comment on the `utils.ConvertToEthBlock` method).

rpc/utils/utils.go
```diff
@@ -174,3 +176,64 @@ func NewRPCTransaction(
...
+		"number":           hexutil.Uint64(block.Round),
+		"gasLimit":         hexutil.Uint64(block.Header.GasLimit),
+		"gasUsed":          hexutil.Uint64(block.Header.GasUsed),
+		"timestamp":        hexutil.Uint64(block.Header.Time),
+		"extraData":        block.Header.Extra,
+		"mixHash":          common.HexToHash(block.Header.MixDigest),
+		"nonce":            ethtypes.EncodeNonce(block.Header.Nonce),
+
+		"uncles":       block.Uncles,
+		"transactions": block.Transactions,
```
ptrus: I think the `block.Transactions` need some conversion as well. The type of these transactions in the response should be `[]*utils.RPCTransaction` (same as the response of the `eth_getTransactionBy*` queries). Or  just a list of`hashes` in case the `fullTx=false`.

rpc/eth/api.go
anonymousGiga: ok

rpc/utils/utils.go
anonymousGiga: ok

rpc/eth/api.go
anonymousGiga: ok

rpc/eth/api.go
anonymousGiga: ok

rpc/utils/utils.go
anonymousGiga: ok

rpc/eth/api.go
anonymousGiga: Done

anonymousGiga:
Please review, thank you.  @ptrus

indexer/utils.go
```diff
@@ -0,0 +1,246 @@
...
+		Round:        number.Uint64(),
+		Header:       innerHeader,
+		Uncles:       []*model.Header{},
+		Transactions: innerTxs,
+	}
+	return innerBlock, nil
+}
+
+func (p *psqlBackend) generateEthBlock(oasisBlock *block.Block, txResults []*client.TransactionWithResults) (*model.Block, error) {
+	bhash, _ := oasisBlock.Header.IORoot.MarshalBinary()
```
ptrus: Let's use:`oasisBlock.Header.EncodedHash()` for the block hash in all cases. As that is the hash of the whole oasis runtime block.

indexer/utils.go
```diff
@@ -0,0 +1,246 @@
...
+		Number:      number,
+		GasLimit:    uint64(defaultGasLimit),
+		GasUsed:     gas,
+		Time:        uint64(block.Header.Timestamp),
+		Extra:       []byte{},
+		MixDigest:   common.Hash{},
+		Nonce:       ethtypes.BlockNonce{},
+		BaseFee:     baseFee,
+	}
+	bhash := ethHeader.Hash()
```
ptrus: Let's use `oasisBlock.Header.EncodedHash()` for the block hash.

Otherwise we will have this issue: https://github.com/starfishlabs/oasis-evm-web3-gateway/issues/56
(the block hash in the response will not match the hash in the request (by which the block is queried))

indexer/utils.go
```diff
@@ -0,0 +1,246 @@
...
+) (*model.Block, error) {
+	bprehash, _ := block.Header.PreviousHash.MarshalBinary()
+	bshash, _ := block.Header.StateRoot.MarshalBinary()
+	bloom := ethtypes.BytesToBloom(ethtypes.LogsBloom(logs))
+
+	var btxhash common.Hash
+	if len(transactions) == 0 {
+		btxhash = EmptyRootHash
+	} else {
+		btxhash = ethtypes.DeriveSha(transactions, trie.NewStackTrie(nil))
```
ptrus: Does any web3 code actually rely on the `btxhash` being commutable from the eth transactions in the block? 
If not, could also use the `block.Header.IORoot` here (as it's done in `main`) but that hash is also dependent on any potential non-eth transactions
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/b7c2545d87037f4642d0650fda572cf68f1b39ba/rpc/utils/utils.go#L35

Which might be problematic if there exists web3 code that relies on how this `txhash` is computed.

storage/psql/psql.go
```diff
@@ -143,3 +159,156 @@ func (db *PostDb) GetContinuesIndexedRound() (uint64, error) {
...
+		"logsBloom":         ethtypes.BytesToBloom(ethtypes.LogsBloom(ethLogs)),
+		"logs":              ethLogs,
+		"transactionHash":   tx.Hash,
+		"gasUsed":           hexutil.Uint64(tx.Gas),
+		"type":              hexutil.Uint64(tx.Type),
+		"blockHash":         tx.BlockHash,
+		"blockNumber":       hexutil.Uint64(tx.Round),
+		"transactionIndex":  hexutil.Uint64(tx.Index),
+		"from":              tx.FromAddr,
+		"to":                tx.ToAddr,
```
ptrus: Apparently some web3 code relies on `to` being explicitly set to `nil` (and not `""`)  for contract creation. Let's handle this as it was handled in: https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/53

storage/psql/psql.go
```diff
@@ -143,3 +159,156 @@ func (db *PostDb) GetContinuesIndexedRound() (uint64, error) {
...
+	}
+
```
ptrus: Seems like `"contractAddress"`  field is missing from the transaction receipt. Same as for `to` address, make sure it explicitly defaults to `nil`, as was done in: https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/58

rpc/utils/utils.go
```diff
@@ -117,3 +119,73 @@ func NewRPCTransaction(
 
 	return resTx, nil
 }
+
+func ConvertToEthBlock(block *model.Block, fullTx bool) map[string]interface{} {
```
ptrus, line 63: This might be nicer if it's defined as a method of the `*model.Block` struct.

```golang
/// ToEthBlock returns a JSON-RPC compatible Ethereum Block from the stored block.
func (b *Block) ToEthBlock(fullTx bool) map[string]interface{} {
....
}
```

rpc/utils/utils.go
```diff
@@ -18,8 +20,8 @@ var (
 	EmptyRootHash        = common.HexToHash("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
 )
 
-// ConvertToEthBlock returns a JSON-RPC compatible Ethereum Block from a given Oasis block and its block result.
-func ConvertToEthBlock(
+// ConstructEthBlock returns a JSON-RPC compatible Ethereum Block from a given Oasis block and its block result.
+func ConstructEthBlock(
```
ptrus: This method is not used by anything anymore and can be removed.

storage/types.go
```diff
@@ -29,6 +29,32 @@ type Storage interface {
...
 	GetContinuesIndexedRound() (uint64, error)
 
+	// GetLastRetainedRound query the minimum round not pruned.
+	GetLastRetainedRound() (uint64, error)
+
 	// GetTransaction queries ethereum transaction by hash.
 	GetTransaction(hash string) (*model.Transaction, error)
+
+	// GetBlockNumber returns the latest block round.
+	GetBlockNumber() (uint64, error)
```
ptrus:
```suggestion
	GetLatestBlockNumber() (uint64, error)
```

rpc/utils/utils.go
```diff
@@ -1,11 +1,13 @@
 package utils
 
 import (
+	"encoding/hex"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
+
```
ptrus: `defaultValidatorAddr` & `EmptyRootHash` are duplicated with `indexer/utils.go`. Maybe create a top-level `common/utils.go` and move those there (and any other shared common functionality)

rpc/eth/api.go
```diff
@@ -625,92 +505,103 @@ func (api *PublicAPI) GetTransactionByBlockNumberAndIndex(blockNum ethrpc.BlockN
 // GetTransactionReceipt returns the transaction receipt by hash.
 func (api *PublicAPI) GetTransactionReceipt(txHash common.Hash) (map[string]interface{}, error) {
 	api.Logger.Debug("eth_getTransactionReceipt", "hash", txHash.Hex())
-	txRef, err := api.backend.QueryTransactionRef(txHash.String())
+	//dbTx, err := api.backend.QueryTransaction(txHash)
```
ptrus: Remove the commented code

rpc/eth/api.go
```diff
@@ -42,6 +38,7 @@ var (
 	ErrBlockNotFound              = errors.New("block not found")
 	ErrTransactionNotFound        = errors.New("transaction not found")
 	ErrTransactionReceiptNotFound = errors.New("transaction receipt not found")
+	ErrOutOfIndex                 = errors.New("out of index")
```
ptrus:
```suggestion
	ErrIndexOutOfRange                 = errors.New("index out of range")
```

storage/psql/psql.go
```diff
@@ -21,12 +26,23 @@ func InitDb(cfg *conf.DatabaseConfig) (*PostDb, error) {
 	if cfg == nil {
 		return nil, errors.New("nil configuration")
 	}
+
+	// parse env var
+	dbUser := os.Getenv("DB_USER")
```
ptrus: This was fixed in https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/68 

All config variables can now also be set via env variables, so special casing these variables here is not needed anymore.

SzeChuanChilliSauce: ok

indexer/utils.go
ptrus: Seems like I was a bit wrong about this, we cannot just return `block.HEader.IORoot` here, as there exists web3-code that relies on this block hash to be `EmptyHash` in case there are no transactions (#69). Probably need to handle the `len(transactions) == 0` case here, and return an `EmptyRootHash` (as it was done previously).

indexer/utils.go
```diff
@@ -0,0 +1,299 @@
...
+	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/client"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
+	"github.com/starfishlabs/oasis-evm-web3-gateway/model"
+)
+
+var (
+	defaultValidatorAddr = "0x0000000000000000000000000000000088888888"
+	defaultGasLimit      = 21000 * 1000
+	EmptyRootHash        = common.HexToHash("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
```
ptrus: Use `ethtypes.EmptyRootHash` instead of re-defining `EmptyRootHash` here.

indexer/backend.go
```diff
@@ -210,16 +123,29 @@ func (p *psqlBackend) Index(
...
+		return err
+	}
+
+	if err := p.storage.Delete(new(model.Log), round); err != nil {
+		return err
+	}
+
+	if err := p.storage.Delete(new(model.Transaction), round); err != nil {
+		return err
+	}
```
ptrus, line 152: I think this should also prune `model.Receipt` and `model.TransactionRef` right?

anonymousGiga: Yes.

anonymousGiga: Done，please reivew it again.

conf/config.go
```diff
@@ -176,5 +176,8 @@ func InitConfig(f string) *Config {
 	err = config.Validate()
 	cobra.CheckErr(err)
 
+	// config.Database.User = os.Getenv("DATABASE__USER")
```
ptrus: This can be removed right?

conf/server.yml
```diff
@@ -11,8 +11,6 @@ database:
   host: "127.0.0.1"
   port: 5432
   db: "postgres"
-  user: "postgres"
```
ptrus: This change can be reverted, so that server.yml serves as a working, example config.

indexer/backend.go
```diff
@@ -286,6 +238,93 @@ func (p *psqlBackend) QueryTransactionRef(hash string) (*model.TransactionRef, e
...
+		"cumulativeGasUsed": hexutil.Uint64(dbReceipt.CumulativeGasUsed),
+		"logsBloom":         ethtypes.BytesToBloom(ethtypes.LogsBloom(ethLogs)),
+		"logs":              ethLogs,
+		"transactionHash":   dbReceipt.TransactionHash,
+		"gasUsed":           hexutil.Uint64(dbReceipt.GasUsed),
+		"type":              hexutil.Uint64(dbReceipt.Type),
+		"blockHash":         dbReceipt.BlockHash,
+		"blockNumber":       hexutil.Uint64(dbReceipt.Round),
+		"transactionIndex":  hexutil.Uint64(dbReceipt.TransactionIndex),
+		"from":              dbReceipt.FromAddr,
```
ptrus: `from`, `to` and `contractAddress` are stored as `string` in the DB. Therefore, these will be `""` in case these are not present (e.g. `to` in the contract creation transactions and `contractAddress` in non-contract creation transactions). 

Let's make these are `null` in case these are not present, as some web3-clients expect these to be `null` and not `""`. For additional info see: https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/53 and https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/58

Either update these in DB to support `nil` (e.g. making the fields `*string`) or handle the `""` case here.

conf/config.go
anonymousGiga: yes

indexer/backend.go
anonymousGiga: OK, here is my previous carelessness, which has been revised now. I have run all the tests now, except that estimategas will fail.

anonymousGiga: The implementation of estimategas（https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/index-more-things/rpc/eth/api.go#L376） is the same as the main branch, and the main branch may have the same problem.    

I test in main branch, it also get the error  ```  err":"invalid argument: missing field","level":"error","method":"/oasis-core.RuntimeClient/Query","module":"grpc/client","msg":"request failed"  ```

anonymousGiga: Please review it，thank you very much. @ptrus

rpc/utils/utils.go
```diff
@@ -1,78 +1,16 @@
...
-		"uncles":          []common.Hash{},
-		"transactions":    transactions,
-		"totalDifficulty": (*hexutil.Big)(big.NewInt(0)),
-	}
-
-	return res, nil
-}
-
 // NewRPCTransaction returns a transaction that will serialize to the RPC representation.
 func NewRPCTransaction(
```
ptrus: `NewRPCTransaction` accepts `blockHash`, `blockNumber` and `index` parameters which I think are not needed anymore now that `model.Transaction` already contains all the fields.

rpc/utils/utils.go
```diff
@@ -129,3 +67,72 @@ func NewRPCTransaction(
...
 	return resTx, nil
 }
+
+func ConvertToEthBlock(block *model.Block, fullTx bool) map[string]interface{} {
+	v1 := big.NewInt(0)
+	diff, _ := v1.SetString(block.Header.Difficulty, 10)
+	transactions := []interface{}{}
+
+	if fullTx {
+		transactions = append(transactions, block.Transactions)
```
ptrus: Should `NewRPCTransaction` be called for each of the transactions here? The transactions in block response should be of the same format as the transactions in the `getTransactionByX` response.

If `NewRPCTransaction` is not needed here, then it also shouldn't be needed in the `getTransactionByX` methods, and can be removed.

indexer/utils.go
```diff
@@ -0,0 +1,304 @@
...
+	// Get convert block, transactions and receipts
+	blk, txs, receipts, err := convertToEthFormat(oasisBlock, ethTxs, logs, txsStatus, results, gasUsed)
+	if err != nil {
+		p.logger.Debug("Failed to ConvertToEthBlock", "height", blockNum, "error", err.Error())
+		return err
+	}
+
+	// Store txs
+	for _, tx := range txs {
+		err = p.storage.Store(tx)
```
pro-wh, line 253: minor question about the storage system: does this quietly replace if the tx is already in the db? if not, we could run into trouble in a case where one of these fails, and we retry, and some of the earlier txs already are stored.

pro-wh (commented):
I've looked through briefly. I think the overall design is fine. And the error handling is pretty nice. Thanks for the improvements and thanks to the other reviewers for their help.

There is a lot of logic in these new, tidied-up oasis <-> eth <-> db structure conversions which I don't have the domain expertise to do a detailed review, but the architecture looks like it'll make it easy for us to make corrections if issues come up.

I see that @ptrus has some suggestions on duplicate information in some function parameters. Please continue with that process.

I asked around on our side, and it seems the plan is to get the linting and testing fixed up separately from this PR. Okay.

rpc/utils/utils.go
SzeChuanChilliSauce: yes, we have solved the problem.

rpc/utils/utils.go
```diff
@@ -121,11 +52,82 @@ func NewRPCTransaction(
 		S:         (*hexutil.Big)(s),
 	}
 
+	blockHash := common.HexToHash(dbTx.BlockHash)
+	txIndex := hexutil.Uint64(dbTx.Index)
 	if blockHash != (common.Hash{}) {
```
ptrus: Is this `if` needed? I don't think `blockHash` will ever be `common.Hash{}`?

storage/psql/psql.go
```diff
@@ -143,3 +144,104 @@ func (db *PostDb) GetContinuesIndexedRound() (uint64, error) {
...
+		Where("tip=?", model.LastRetained).
+		Select()
+	if err != nil {
+		return 0, err
+	}
+
+	return retainedRound.Round, nil
+}
+
+func (db *PostDb) GetLatestBlockNumber() (uint64, error) {
```
ptrus, line 182: Please add documentation comment (and also all other added methods in this file)

rpc/utils/utils.go
ptrus: Please add documentation comment

rpc/utils/utils.go
```diff
@@ -121,11 +52,82 @@ func NewRPCTransaction(
...
+		"transactions":     transactions,
+		"hash":             common.HexToHash(block.Hash),
+		"size":             hexutil.Uint64(len(serialized)),
+		"totalDifficulty":  (*hexutil.Big)(big.NewInt(0)),
+	}
+
+	return res
+}
+
+func Db2EthLogs(dbLogs []*model.Log) []*ethtypes.Log {
```
ptrus, line 105: Please add documentation comment

rpc/eth/api.go
```diff
@@ -803,14 +576,14 @@ func (api *PublicAPI) GetBlockHash(blockNum ethrpc.BlockNumber, _ bool) (common.
...
 func (api *PublicAPI) BlockNumber() (hexutil.Uint64, error) {
 	api.Logger.Debug("eth_getBlockNumber start")
-	blk, err := api.client.GetBlock(api.ctx, client.RoundLatest)
+	blockNumber, err := api.backend.BlockNumber()
 	if err != nil {
-		api.Logger.Error("eth_getBlockNumber get the latest number", "error", err.Error())
+		api.Logger.Error("failed to get the latest block number", "err:", err)
 		return 0, err
 	}
+	api.Logger.Debug("eth_getBlockNumber get the current number", "blockNumber:", blockNumber)
```
ptrus: Please remove the `:` in log fields (e.g. just `"err"`/`"blockNumber"` instead of `"err:"`/`"blockNumber:"`), since the logger should already format these fields nicely.

Also elsewhere.

rpc/eth/api.go
```diff
@@ -744,55 +538,34 @@ func (api *PublicAPI) GetLogs(filter filters.FilterCriteria) ([]*ethtypes.Log, e
...
-			}
-			return nil, fmt.Errorf("get block %d: %w", round, err)
-		}
-		_, _, _, blockLogs, err := api.getRPCBlockData(block, true)
-		if err != nil {
-			return nil, fmt.Errorf("convert block %d to rpc block: %w", block.Header.Round, err)
-		}
-
-		// TODO: filter addresses and topics
+	//var ethLogs []*ethtypes.Log
```
ptrus: Remove this unsued commented code.

rpc/eth/api.go
```diff
@@ -744,55 +538,34 @@ func (api *PublicAPI) GetLogs(filter filters.FilterCriteria) ([]*ethtypes.Log, e
...
+	//for round := startRoundInclusive; ; /* see explicit break */ round++ {
+	//	dbLogs, err := api.backend.GetLogs(*filter.BlockHash)
+	//	if err != nil {
+	//		api.Logger.Error("get logs, err:", err)
+	//		continue
+	//	}
+	//	// TODO: filter addresses and topics
```
ptrus: Keep this `TODO: filter addresses and topics` as a note that filtering by addresses/topics is currently not supported.

rpc/eth/api.go
```diff
@@ -93,101 +83,24 @@ func (api *PublicAPI) roundParamFromBlockNum(blockNum ethrpc.BlockNumber) (uint6
 	case ethrpc.LatestBlockNumber:
 		return client.RoundLatest, nil
 	case ethrpc.EarliestBlockNumber:
-		lrBlk, err := api.client.GetLastRetainedBlock(api.ctx)
+		var earliest uint64 = 0
```
ptrus:
```suggestion
		var earliest uint64
```

indexer/utils.go
```diff
@@ -0,0 +1,304 @@
...
+			Removed:   log.Removed,
+		}
+
+		res = append(res, dbLog)
+	}
+
+	return res
+}
+
+func ConvertHashToString(hash common.Hash) string {
```
ptrus: I think this function isn't used anywhere and can be removed.

rpc/utils/utils.go
SzeChuanChilliSauce: ok, it has been removed.

SzeChuanChilliSauce: ok, it has been removed.

indexer/utils.go
ptrus: Nice catch, this would currently fail with:
```
ERROR #23505 duplicate key value violates unique constraint
```

Since the way the indexer is currently designed it might happen that inserting for same round will be retried (e.g. if the process is restarted in the middle of indexing an round), the `storage.Store` method should be updated to do an `"upsert"` (e.g. add `ON CONFLICT DO UPDATE` to the insert query)

ptrus (commented):
Thanks this looks quite good, mostly minor stylistic comments. 

The `storage.Store()` should be updated to do an "upsert", see comment for details.

rpc/eth/api.go
SzeChuanChilliSauce: ok

indexer/utils.go
ptrus: Thanks for the fixes. @SzeChuanChilliSauce  This is the last one, after it this is good to merge :+1:

SzeChuanChilliSauce: ok, we'll fix it as soon as possible.

storage/psql/psql.go
```diff
@@ -69,23 +70,31 @@ func (db *PostDb) GetTransaction(hash string) (*model.Transaction, error) {
 	return tx, nil
 }
 
-// Store stores data.
-func (db *PostDb) Store(value interface{}) error {
-	_, err := db.Db.Model(value).Insert()
+// insertOrUpdate updates record when it exists, or inserts.
+func (db *PostDb) insertOrUpdate(value interface{}) error {
+	query := db.Db.Model(value).WherePK()
```
ptrus, line 88: I think this might be more idiomatic and also safer since this operation is atomic: 
 
```golang
func (db *PostDb) upsert(value interface{}) error {
        // PKs are required for ON CONFLICT DO UPDATE
	pks := db.Db.Model(value).Table().TableModel().Table().PKs
	b := ""
	for i, f := range pks {
		if i > 0 {
			b += ","
		}
		b += string(f.Column)
	}
	_, err := db.Db.Model(value).
		OnConflict(fmt.Sprintf("(%s) DO UPDATE", b)).
		Insert()
	return err
}
```

The only downside/ugly part is that we need to load the PKs (could probably just use the first one). 


Anyway feel free to update or keep it as it is. Hopefully `bun` (used in https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/74) has a nicer way to do this :)

ptrus (approved):
Thanks, looks good!

One (optional) comment, feel free to decide yourself if you want to fix it or keep it as it is.

Maybe also cleanup commits a bit (could also squash in a single one if you prefer). Other than that, feel free to merge!

storage/psql/psql.go
anonymousGiga: We tried this method and found that there were some problems to be solved, so we merged this branch first and then prepared to solve it in #74.

PR #53, opened 2021-11-11
eth_getTransactionReceipt: return null instead of empty address

ptrus:
This makes the transaction receipt response compatible with `ropsten` (and other networks?), where in contract creation transactions the `to` field should is  set to `null` instead of an empty string.
image

### DIFF
--- a/rpc/eth/api.go
+++ b/rpc/eth/api.go
@@ -668,20 +668,26 @@ func (api *PublicAPI) GetTransactionReceipt(txHash common.Hash) (map[string]inte
 		"blockHash":         txRef.BlockHash,
 		"blockNumber":       hexutil.Uint64(txRef.Round),
 		"transactionIndex":  hexutil.Uint64(txRef.Index),
-		"from":              ethTx.FromAddr,
-		"to":                ethTx.ToAddr,
+		"from":              nil,
+		"to":                nil,
+	}
+	if ethTx.FromAddr != "" {
+		receipt["from"] = ethTx.FromAddr
+	}
+	if ethTx.ToAddr != "" {
+		receipt["to"] = ethTx.ToAddr
 	}
 	if logs == nil {
 		receipt["logs"] = [][]*ethtypes.Log{}
 	}
-	if len(ethTx.ToAddr) == 0 && txResults[txRef.Index].Result.IsSuccess() {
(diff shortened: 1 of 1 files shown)


rpc/eth/api.go
```diff
@@ -668,20 +668,24 @@ func (api *PublicAPI) GetTransactionReceipt(txHash common.Hash) (map[string]inte
...
 		"blockNumber":       hexutil.Uint64(txRef.Round),
 		"transactionIndex":  hexutil.Uint64(txRef.Index),
-		"from":              ethTx.FromAddr,
-		"to":                ethTx.ToAddr,
+	}
+	if ethTx.FromAddr != "" {
+		receipt["from"] = ethTx.FromAddr
+	}
+	if ethTx.ToAddr != "" {
+		receipt["to"] = ethTx.ToAddr
```
kostko, line 678: Will this result in the correct behavior or should it be explicitly set to `nil`?

kostko (approved):

rpc/eth/api.go
ptrus: Added explicit `nil`s, the response should now match ropsten response:
```
{"jsonrpc":"2.0","id":1,"result":{"blockHash":"0xae29df20d690e2b2bfefae9b4f4191b40e940bdb8a4159c345303c5eaba1d5e9","blockNumber":"0x5","contractAddress":"0xfbcb580dd6d64fbf7caf57fb0439502412324179","cumulativeGasUsed":"0x2fd3d","from":"0x90adE3B7065fa715c7a150313877dF1d33e777D5","gasUsed":"0x2fd3d","logs":[],"logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","status":"0x1","to":null,"transactionHash":"0x655faf929323d922609a90876494411b3257e222632af23d5f73148bf36a6e58","transactionIndex":"0x0","type":"0x0"}}
```

kostko (approved):

Issue #55, opened 2021-11-11
`eth_getBlockByHash` & `eth_getBlockByNumber`: missing transaction blockHash", "blockNumber" and "from" fields


ptrus:
If `fullTx` parameter is set to `true` (currently assumed always is: https://github.com/starfishlabs/oasis-evm-web3-gateway/issues/54) all block transactions are included in the `transactions` field. In oasis-evm-web3-gateway case some transaction fields are missing.

The transactions included in the block data should match the format of the `eth_getTransaction` queries, the following fields are expected:
```
    "blockHash":"0x1d59ff54b1eb26b013ce3cb5fc9dab3705b415a67127a003c3e61eb445bb8df2",
    "blockNumber":"0x5daf3b", // 6139707
    "from":"0xa7d9ddbe1f17865597fbd27ec712455208b6b76d",
    "gas":"0xc350", // 50000
    "gasPrice":"0x4a817c800", // 20000000000
    "hash":"0x88df016429689c079f3b2f6ad39fa052532c56795b733da78a91ebe6a713944b",
    "input":"0x68656c6c6f21",
    "nonce":"0x15", // 21
    "to":"0xf02c1c8e6114b1dbe8937a39260b5b0a374432bb",
    "transactionIndex":"0x41", // 65
    "value":"0xf3dbb76162000", // 4290000000000000
    "v":"0x25", // 37
    "r":"0x1b5e176d927f8e9ab405058b2d2457392da3e20f328b16ddabcebc33eaac5fea",
    "s":"0x4ba69724e8f69de52f0125ad8b3c5c2cef33019bac3249e2c0a2192766d1721c"
```

In oasis-evm-web3-gateway case, the following fields for a transaction are missing: `blockHash`, `blockNumber`, `from`, `transactionIndex`. (note: the `eth_getTransaction` response does contain these fields, so this is only problematic in the `block` queries).

Issue #54, opened 2021-11-11
`eth_getBlockByHash` & `eth_getBlockByNumber` ignore the `fullTx` parameter


ptrus:
`eth_getBlockByHash` & `eth_getBlockByNumber` ignore the 2nd paramter which is:
> Boolean - If true it returns the full transaction objects, if false only the hashes of the transactions.

Seems like currently the requests always return full transactions.

PR #61, opened 2021-11-12
Starfish/hotfix/full tx parameter problem

SzeChuanChilliSauce:
issue：#54 #55

### DIFF
--- a/rpc/eth/api.go
+++ b/rpc/eth/api.go
@@ -144,15 +144,17 @@ func (api *PublicAPI) getRPCBlockData(oasisBlock *block.Block, fullTx bool) (uin
 			hash := ethTx.Hash()
 			ethRPCTxs = append(ethRPCTxs, hash)
 		} else {
-			blockHash := common.HexToHash(oasisBlock.Header.EncodedHash().Hex())
-			txIndex := hexutil.Uint64(txIndex)
-			ethRPCTx, err := utils.NewRPCTransaction2(ethTx, blockHash, oasisBlock.Header.Round, txIndex)
+			dbTx, err := api.backend.QueryTransaction(ethTx.Hash())
 			if err != nil {
-				api.Logger.Error("Failed to create RPCTransaction in getRPCBlockData", "height", blockNum, "index", txIndex, "error", err.Error())
+				api.Logger.Error("Failed to QueryTransaction", "hash", ethTx.Hash(), "error", err.Error())
 				continue
 			}
-
-			ethRPCTxs = append(ethRPCTxs, ethRPCTx)
+			rpcTx, err := api.getRPCTransaction(dbTx)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


rpc/eth/api.go
```diff
@@ -104,11 +104,12 @@ func (api *PublicAPI) roundParamFromBlockNum(blockNum ethrpc.BlockNumber) (uint6
...
 }
 
-func (api *PublicAPI) getRPCBlockData(oasisBlock *block.Block) (uint64, ethtypes.Transactions, uint64, []*ethtypes.Log, error) {
+func (api *PublicAPI) getRPCBlockData(oasisBlock *block.Block, fullTx bool) (uint64, []interface{}, uint64, []*ethtypes.Log, error) {
 	encoded := oasisBlock.Header.EncodedHash()
 	bHash, _ := encoded.MarshalBinary()
 	blockNum := oasisBlock.Header.Round
-	ethTxs := ethtypes.Transactions{}
+	// ethTxs := ethtypes.Transactions{}
+	ethRPCTxs := []interface{}{}
```
ptrus, line 112: Add a comment that `ethRPCTxs` will contain either `RPCTransactions` or the transaction hashes, depending on the `fullTx` parameter.

rpc/eth/api.go
```diff
@@ -135,7 +136,21 @@ func (api *PublicAPI) getRPCBlockData(oasisBlock *block.Block) (uint64, ethtypes
 		}
 
 		gasUsed += uint64(ethTx.Gas())
-		ethTxs = append(ethTxs, ethTx)
+
+		if !fullTx {
+			hash := ethTx.Hash()
+			ethRPCTxs = append(ethRPCTxs, hash)
+		} else {
+			blockHash := common.HexToHash(oasisBlock.Header.EncodedHash().Hex())
```
ptrus, line 144: Could `RPCTransaction` be loaded from the database here?
```
		dbTx, err := api.backend.QueryTransaction(ethTx.Hash())
		if err != nil {
			api.Logger.Error("Failed to QueryTransaction", "hash", ethTx.Hash(), "error", err.Error())
			continue
		}
		rpcTx, err := api.getRPCTransaction(dbTx)
		if err != nil {
			api.Logger.Error("Failed to getRPCTransaction", "hash", ethTx.Hash(), "error", err.Error())
			continue
		}
```

That way, you don't need the added `NewRPCTransaction2` code, which will in any case be only temporary as this whole part should be removed in https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/44 (and replaced by loading everything from the db).

ptrus (commented):
Left some minor comments

rpc/eth/api.go
SzeChuanChilliSauce: ok，We modified this code.

ptrus (approved):

ptrus:
looks good!

PR #78, opened 2021-11-25
Modify nil to [] for getLogs

### LINKED ISSUE #77: If the return value does not exist, can null be changed to "" or [] to adapt to the rust interface?
{
	"jsonrpc":"2.0",
	"method":"eth_getLogs",
	"params":[
		{ "fromBlock": "0x200", "toBlock" : "0x205"}
	],
	"id":3
}
--------------------------------------
{
    "jsonrpc": "2.0",
    "id": 3,
    "result": null --> []
}

anonymousGiga:
Fix #77


rpc/eth/api.go
```diff
@@ -513,12 +513,13 @@ func (api *PublicAPI) GetTransactionReceipt(txHash common.Hash) (map[string]inte
...
 	api.Logger.Debug("eth_getLogs", "filter", filter)
 
+	ethLogs := []*ethtypes.Log{}
 	startRoundInclusive := client.RoundLatest
 	endRoundInclusive := client.RoundLatest
 	if filter.BlockHash != nil {
 		round, err := api.backend.QueryBlockRound(*filter.BlockHash)
 		if err != nil {
-			return nil, fmt.Errorf("query block round: %w", err)
+			return ethLogs, fmt.Errorf("query block round: %w", err)
```
kostko: Can you explain the reasoning behind this change? If there's an error then `nil` seems like the correct thing to return? And you do return `nil` in case of other errors below?

SzeChuanChilliSauce: shoule we treat "data not found" as an error? if data not found, we should return [ ]. if there are other errors, nil should be returned. @anonymousGiga

kostko: But this returns both, an error and an empty list? Should this then only return an empty list with nil error?

anonymousGiga: If the hash value is not empty here and the corresponding information cannot be found from the database, it is not an error for the user. In this case, it should return []. Maybe we can judge the error type directly in backend.queryblockround without modifying here.

anonymousGiga: Sorry， it seems that can't judge in backend.queryblockround. I'll add error type judgment here.

SzeChuanChilliSauce (approved):

SzeChuanChilliSauce (approved):

SzeChuanChilliSauce (approved):

ptrus:
I think this PR is outdated / not needed.

PR #82, opened 2021-11-29
various fixes

### LINKED ISSUE #63: nit: ethrpc.BlockNum is int64, zero and negatives used for special values
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/22eb39c828e9319dbbb9e14f9fe1be9192f8715f/rpc/eth/api.go#L103

golang doesn't have enums nor a way to warn if enums aren't all checked a la gcc's `-Wswitch-enum` compiler flag, and go-ethereum/rpc/types.go defines `PendingBlockNumber`, `LatestBlockNumber`, and `EarliestBlockNumber` as -2, -1, and 0 resp special case values.  It's unclear whether go-ethereum would ever change this API, but if they added a new negative special value, the `rpc/eth/api.go` code will silently convert it into a very large 64-bit integer, since golang's conversion does not range check and panic for us.

since there is no `-Wswitch-enum`, a reasonable defense-in-depth approach would be to check for negative numbers and panic, along with a comment so that future engineers who encounter the panic will get a hint as to what needs to be done.  (i don't think a compile-time test to see if go-ethereum has extended the `const` list is feasible; ideas welcome!)

### LINKED ISSUE #64: nit: misleading error message
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/22eb39c828e9319dbbb9e14f9fe1be9192f8715f/rpc/eth/api.go#L180

this looks cut-n-pasted.  here it is `roundParamFromBlockNum` that failed, not `GetBlock`.

### LINKED ISSUE #65: nit: misleading error message
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/22eb39c828e9319dbbb9e14f9fe1be9192f8715f/rpc/eth/api.go#L198

this looks cut-n-pasted. here it is `roundParamFromBlockNum` that failed, not `Get Transactions`.

also, it is probably a good idea to maximize greppability of log messages to say `GetTransactions` on L204, instead of `Get Transactions`, but this is very minor.

### LINKED ISSUE #66: nit: misleading error message
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/22eb39c828e9319dbbb9e14f9fe1be9192f8715f/rpc/eth/api.go#L218

here it is not `GetBlock` which failed, but `roundParamFromBlockNum` that failed.

### LINKED ISSUE #67: tiny nit: indistinguishable error messages
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/22eb39c828e9319dbbb9e14f9fe1be9192f8715f/rpc/eth/api.go#L243

https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/22eb39c828e9319dbbb9e14f9fe1be9192f8715f/rpc/eth/api.go#L303

having (cut-n-pasted) error strings that are pretty much indistinguishable makes tracking down errors more difficult, since the `error` interface are (mostly) strings and in our usage does not include a stack backtrace for the initial error that caused the cascade.  future programmers who encounter such strings in logs will grep for the text and find multiple potential locations where the error might have originated.  ideally every error string is uniquely greppable or otherwise has enough context.

### LINKED ISSUE #73: Add test cases for recently fixed issues
Once tests and CI is fixed (https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/45) add tests for recently fixed issues:
- [x] test correct `transaction.input` in `getTransaction` responses: https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/71
- [x] test to ensure `EmptyRootHash` is returned for transactionRoot in case no eth transactions in block https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/70 (already tested)
- [x] test to ensure `fullTx` parameter is handled in `getBlock*` requests (https://github.com/starfishlabs/oasis-evm-web3-gateway/issues/54)
- [x] ensure expected `transactions` output in `getBlock*` requests (https://github.com/starfishlabs/oasis-evm-web3-gateway/issues/55)
- [x] ensure explicit `null` in transaction receipt response (https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/58 https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/53)

### LINKED ISSUE #81: Indexer: storing logs
Indexing logs fails with: 
```
{"caller":"utils.go:242","err":"pg: Model(non-pointer []*model.Log)","height":1038,"index":0,"level":"error","logs":[{"Address":"0xaed2d02cb1f5195cfe39f84c56c4c8788460f6a3","Topics":["0xbc7cd75a20ee27fd9adebab32041f755214dbc6bffa90cc0225b39da2e5c2d3b","0x000000000000000000000000f774fbc4fbafd547579cd4f407b5ef6a7a55fca4"],"Data":""}],"module":"indexer","msg":"Failed to store logs","ts":"2021-11-26T22:25:56.711711804Z"}
```

It seems that the error is that the `storage.Store` method expects a single Model value and not an array of values, as is the case with logs here:
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/496d4baeb13c7d4e9096f6225742677eb7638ee2/indexer/utils.go#L241-L244
The easiest solution is probably to just iterate through the logs and store them. There might be better solutions.

Additionally, failing to store logs should fail with an error and terminate the process (e.g. change the `continue` in the above code into returning an error).

The fix PR should also include a unit test for this case.

ptrus:
- [x] `spinup-oasis-stack.sh` fail on `jq` errors (e.g. if `EMERALD_VERSION_NOT_SET` is not set)
- [x] fix handling `"latest"` round in block queries
- [x] fix not loading block transactions 
- [x] fix indexing logs https://github.com/starfishlabs/oasis-evm-web3-gateway/issues/81 
- [x] tests for recently fixed issues: https://github.com/starfishlabs/oasis-evm-web3-gateway/issues/73

### DIFF
--- a/tests/rpc/rpc_test.go
+++ b/tests/rpc/rpc_test.go
@@ -169,6 +169,25 @@ func TestEth_GetBlockByNumberAndGetBlockByHash(t *testing.T) {
 	require.Equal(t, number, blk3.Number())
 }
 
+func TestEth_GetBlockByNumberLatest(t *testing.T) {
+	ctx := context.Background()
+	ec := localClient()
+
+	// Explicitly query latest block number.
+	block, err := ec.BlockByNumber(ctx, nil)
+	require.NoError(t, err, "get latest block number")
+	require.Greater(t, block.NumberU64(), uint64(0))
+}
+
+func TestEth_GetBlockTransactionCountByNumberLatest(t *testing.T) {
+	ctx := context.Background()
+	ec := localClient()
+
+	// Explicitly query latest block number.
+	_, err := ec.PendingTransactionCount(ctx)
+	require.NoError(t, err, "get pending(=latest) transaction count")
+}
+
 func TestEth_BlockNumber(t *testing.T) {
 	ec := localClient()
 	ctx := context.Background()
--- a/tests/rpc/rpc_test.go
+++ b/tests/rpc/rpc_test.go
@@ -234,20 +234,7 @@ func TestEth_GetTransactionByHash(t *testing.T) {
 	require.Equal(t, uint64(1), receipt.Status)
 	require.NotNil(t, receipt)
 
-	// go-ethereum's Transaction struct always computes transaction hash on-the-fly
-	// instead of simply returning the hash from TransactionBy* API responses.
-	// To overcome this, we perform getTransactionByHash query with raw HTTP
-	// client and use the tx's hash from that response.
-	// For details, see https://github.com/starfishlabs/oasis-evm-web3-gateway/issues/72
-	rpcRes := call(t, "eth_getTransactionByHash", []string{receipt.TxHash.Hex()})
-
-	tx2 := make(map[string]interface{})
-	rpcErr := json.Unmarshal(rpcRes.Result, &tx2)
-	require.NoError(t, rpcErr)
-	require.NotNil(t, tx2)
-	require.Equal(t, signedTx.Hash(), common.HexToHash(tx2["hash"].(string)))
-
-	tx3, _, err := ec.TransactionByHash(ctx, receipt.TxHash)
+	tx2, _, err := ec.TransactionByHash(ctx, receipt.TxHash)
 	require.NoError(t, err)
-	require.NotNil(t, tx3)
+	require.NotNil(t, tx2)
 }
--- a/tests/rpc/rpc_test.go
+++ b/tests/rpc/rpc_test.go
@@ -61,6 +61,46 @@ func call(t *testing.T, method string, params interface{}) *Response {
 	return rpcRes
 }
 
+func submitTransaction(ctx context.Context, t *testing.T, to common.Address, amount *big.Int, gasLimit uint64, gasPrice *big.Int, data []byte) *types.Receipt {
+	ec := localClient()
+	chainID, err := ec.ChainID(context.Background())
+	require.NoError(t, err)
+
+	nonce, err := ec.NonceAt(context.Background(), oasisTesting.Dave.EthAddress, nil)
+	require.Nil(t, err, "get nonce failed")
+
+	// Create transaction
+	tx := types.NewTransaction(
+		nonce,
+		to,
+		amount,
+		gasLimit,
+		gasPrice,
+		data,
+	)
+	signer := types.LatestSignerForChainID(chainID)
+	signature, err := crypto.Sign(signer.Hash(tx).Bytes(), daveKey)
+	require.Nil(t, err, "sign tx")
+
+	signedTx, err := tx.WithSignature(signer, signature)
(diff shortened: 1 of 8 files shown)


kostko (approved):

tests/rpc/rpc_test.go
```diff
@@ -169,66 +192,112 @@ func TestEth_GetBlockByNumberAndGetBlockByHash(t *testing.T) {
...
-	require.Nil(t, err, "sign tx")
+	receipt := submitTransaction(ctx, t, to, big.NewInt(1), 3000003, big.NewInt(2), data)
+	require.EqualValues(t, 1, receipt.Status)
+	require.NotNil(t, receipt)
 
-	signedTx, err := tx.WithSignature(signer, signature)
-	require.Nil(t, err, "pack tx")
+	tx2, _, err := ec.TransactionByHash(ctx, receipt.TxHash)
+	require.NoError(t, err)
+	require.NotNil(t, tx2)
```
matevz, line 242: Hm, you removed the check which compares transaction hashes and fails for me:
```
	require.Equal(t, tx2.Hash(), receipt.TxHash)
```

Or am I comparing two different things?

tests/tools/spinup-oasis-stack.sh
```diff
@@ -11,9 +11,8 @@ set -ue
 # - OASIS_NODE_DATADIR: path to temprorary oasis-node data dir e.g. /tmp/eth-runtime-test
 
 function emerald_ver {
-	echo $OASIS_EMERALD_VERSION | cut -d \- -f 1 | cut -d + -f 1 | cut -d . -f $1
+  echo $OASIS_EMERALD_VERSION | cut -d \- -f 1 | cut -d + -f 1 | cut -d . -f $1
```
matevz, line 14: Do we have a linter for bash?

matevz (changes requested):
Some minor nits.

tests/tools/spinup-oasis-stack.sh
ptrus: I have a local one configured

ptrus: I can open an issue to add it, but probably not needed while there is only this single script.

tests/rpc/rpc_test.go
ptrus: Fixed in 7a1a8eafc9a88ae527c60be191e695279b8e9bae

tests/tools/spinup-oasis-stack.sh
matevz: Let's just leave it for now and we'll add bash linting once we have at least two bash scripts :)

matevz (approved):
Great job! :shipit:

PR #83, opened 2021-11-29
fix storing logs.

SzeChuanChilliSauce:
issue: #81


ptrus:
fixed in https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/82

Issue #96, opened 2021-12-07
Add more database indexes where needed


kostko:
All the database models and queries should be reviewed and appropriate indexes added where needed. One such example is the `blocks` table which seems to lack an index on `round` even though there are many round-related queries.

ptrus:
Should be done after https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/74 in case switching the library changes the way the db is initialized

ptrus:
Obvious ones that are missing

Transactions:
- on `block_hash`
- on `round` 

Blocks:
- on `round`

Logs:
- on `round`
- on `block_hash`

PR #101, opened 2021-12-07
replace pg with bun

SzeChuanChilliSauce:
new branch of replacing pg with bun.


SzeChuanChilliSauce:
i checked out a new branch with the latest commit of main branch.

model/db.go
```diff
@@ -57,11 +57,11 @@ type Transaction struct {
...
 // Block represents ethereum block.
 type Block struct {
-	Hash         string `pg:",pk"`
-	Round        uint64 `pg:",use_zero"`
+	Hash         string `bun:",pk"`
+	Round        uint64
 	Header       *Header
 	Uncles       []*Header
-	Transactions []*Transaction `pg:"rel:has-many"`
+	Transactions []*Transaction
```
ptrus, line 64: Why was the relation removed here? I think this should something like:

```golang
Transactions []*Transaction `bun:"rel:has-many,join:id=hash"`
```

Similar for all other table relations

ptrus: Here is the documentation: https://bun.uptrace.dev/guide/relations.html

storage/psql/psql.go
```diff
@@ -210,58 +212,59 @@ func (db *PostDB) GetBlockByHash(blockHash string) (*model.Block, error) {
...
-	return len(blk.Transactions), nil
+	return len(block.Transactions), nil
 }
 
 // GetBlockTransaction returns transaction by bock hash and transaction index.
 func (db *PostDB) GetBlockTransaction(blockHash string, txIndex int) (*model.Transaction, error) {
-	blk := new(model.Block)
-	err := db.DB.Model(blk).Where("hash=?", blockHash).Relation("Transactions").Select()
+	block := new(model.Block)
+	err := db.DB.NewSelect().Model(block).Where("hash = ?", blockHash).Scan(context.Background())
```
ptrus, line 249: I think you still need the `.Relation("Transactions")` similar as it was before. Since now that you've set up the relation it won't be loaded automatically. See example: https://bun.uptrace.dev/guide/getting-started.html#table-relations

storage/psql/psql.go
```diff
@@ -210,58 +212,59 @@ func (db *PostDB) GetBlockByHash(blockHash string) (*model.Block, error) {
...
-	return len(blk.Transactions), nil
+	return len(block.Transactions), nil
 }
 
 // GetBlockTransactionCountByHash returns the count of transactions in block by block hash.
 func (db *PostDB) GetBlockTransactionCountByHash(blockHash string) (int, error) {
-	blk := new(model.Block)
-	err := db.DB.Model(blk).Where("hash=?", blockHash).Relation("Transactions").Select()
+	block := new(model.Block)
+	err := db.DB.NewSelect().Model(block).Where("hash = ?", blockHash).Scan(context.Background())
```
ptrus, line 238: I think you need to add the `.Relation("Transactions")` similar as it was before. Since now that you've set up the relation, it won't be loaded automatically. See example: https://bun.uptrace.dev/guide/getting-started.html#table-relations

storage/psql/psql.go
```diff
@@ -210,58 +212,59 @@ func (db *PostDB) GetBlockByHash(blockHash string) (*model.Block, error) {
...
-	return blk, nil
+	return block, nil
 }
 
 // GetBlockTransactionCountByNumber returns the count of transactions in block by block number.
 func (db *PostDB) GetBlockTransactionCountByNumber(round uint64) (int, error) {
-	blk := new(model.Block)
-	err := db.DB.Model(blk).Where("round=?", round).Relation("Transactions").Select()
+	block := new(model.Block)
+	err := db.DB.NewSelect().Model(block).Where("round = ?", round).Scan(context.Background())
```
ptrus, line 227: I think you need to add the `.Relation("Transactions")` similar as it was before. Since now that you've set up the relation, it won't be loaded automatically. See example: https://bun.uptrace.dev/guide/getting-started.html#table-relations

storage/psql/psql.go
```diff
@@ -210,58 +212,59 @@ func (db *PostDB) GetBlockByHash(blockHash string) (*model.Block, error) {
 
 // GetBlockByNumber returns the block for the given round.
 func (db *PostDB) GetBlockByNumber(round uint64) (*model.Block, error) {
-	blk := new(model.Block)
-	err := db.DB.Model(blk).Where("round=?", round).Relation("Transactions").Select()
+	block := new(model.Block)
+	err := db.DB.NewSelect().Model(block).Where("round = ?", round).Scan(context.Background())
```
ptrus, line 216: I think you need to add the `.Relation("Transactions")` similar as it was before. Since now that you've set up the relation, it won't be loaded automatically. See example: https://bun.uptrace.dev/guide/getting-started.html#table-relations

storage/psql/psql.go
```diff
@@ -188,19 +190,19 @@ func (db *PostDB) GetLastRetainedRound() (uint64, error) {
...
 
-	return blk.Round, nil
+	return block.Round, nil
 }
 
 // GetBlockByHash returns the block for the given hash.
 func (db *PostDB) GetBlockByHash(blockHash string) (*model.Block, error) {
 	blk := new(model.Block)
-	err := db.DB.Model(blk).Where("hash=?", blockHash).Relation("Transactions").Select()
+	err := db.DB.NewSelect().Model(blk).Where("hash = ?", blockHash).Scan(context.Background())
```
ptrus, line 205: I think you need to add the `.Relation("Transactions")` similar as it was before. Since now that you've set up the relation, it won't be loaded automatically. See example: https://bun.uptrace.dev/guide/getting-started.html#table-relations

storage/psql/psql.go
```diff
@@ -210,58 +212,59 @@ func (db *PostDB) GetBlockByHash(blockHash string) (*model.Block, error) {
...
 
-	return blk.Transactions[txIndex], nil
+	return block.Transactions[txIndex], nil
 }
 
 // GetTransactionReceipt returns receipt by transaction hash.
 func (db *PostDB) GetTransactionReceipt(txHash string) (*model.Receipt, error) {
 	receipt := new(model.Receipt)
-	if err := db.DB.Model(receipt).Where("transaction_hash=?", txHash).Select(); err != nil {
+	err := db.DB.NewSelect().Model(receipt).Where("transaction_hash = ?", txHash).Scan(context.Background())
```
ptrus, line 266: I think you need to add the `.Relation("Logs")` similar as it was before. Since now that you've set up the relation, it won't be loaded automatically. See example: https://bun.uptrace.dev/guide/getting-started.html#table-relations

storage/psql/psql.go
```diff
@@ -71,6 +79,31 @@ func (db *PostDB) GetTransaction(hash string) (*model.Transaction, error) {
 }
 
 // upsert updates record when PK conflicts, otherwise inserts.
+func (db *PostDB) upsertSingle(value interface{}) error {
+	exists, err := db.DB.NewSelect().Model(value).Exists(context.Background())
+	if err != nil && err != sql.ErrNoRows {
+		return err
+	}
+	if exists {
```
ptrus, line 87: I think the above check is not needed. Just doing:

```golang
		// PKs are required for ON CONFLICT DO UPDATE
		typ := reflect.TypeOf(value)
		table := db.DB.Table(typ)
		pks := make([]string, len(table.PKs))
		for i, f := range table.PKs {
			pks[i] = f.Name
		}
		_, err = db.DB.NewInsert().
			Model(value).
			On(fmt.Sprintf("CONFLICT (%v) DO UPDATE", strings.Join(pks, ","))).
			Exec(context.Background())
```

should do the correct thing if the record exists or if it doesn't.

storage/psql/psql.go
```diff
@@ -2,54 +2,64 @@ package psql
...
+	return &PostDB{DB: db}, nil
 }
 
 // GetTransactionRef returns block hash, round and index of the transaction.
 func (db *PostDB) GetTransactionRef(txHash string) (*model.TransactionRef, error) {
 	tx := new(model.TransactionRef)
-	err := db.DB.Model(tx).
-		Where("eth_tx_hash=?", txHash).
-		Select()
+	err := db.DB.NewSelect().Model(tx).Where("eth_tx_hash = ? ", txHash).Scan(context.Background())
```
ptrus, line 62: All methods should be updated to accept `context.Background()` as a parameter and using here, instead of using `contect.Background()`.

For example, `GetgTransactionRef` should be updated to:
`GetTransactionRef(ctx context.Context, txHash string)`

This can also be done in a seperate PR.

storage/psql/psql.go
```diff
@@ -2,54 +2,64 @@ package psql
...
+		pgdriver.WithUser(cfg.User),
+		pgdriver.WithPassword(cfg.Password),
+		pgdriver.WithTLSConfig(nil),
+		pgdriver.WithDialTimeout(time.Duration(cfg.DialTimeout)*time.Second),
+		pgdriver.WithReadTimeout(time.Duration(cfg.ReadTimeout)*time.Second),
+		pgdriver.WithWriteTimeout(time.Duration(cfg.WriteTimeout)*time.Second))
+
+	// open
+	sqlDB := sql.OpenDB(pgConn)
+	maxOpenConns := 4 * runtime.GOMAXPROCS(0)
```
ptrus, line 43: Add a config parameter for `maxOpenConns` and in case it's `0` use `4 * runtime.GOMAXPROCS(0)` as a default

storage/psql/psql.go
ptrus: Also for `Uncles`.

model/db.go
```diff
@@ -57,30 +57,31 @@ type Transaction struct {
...
 // Block represents ethereum block.
 type Block struct {
-	Hash         string `pg:",pk"`
-	Round        uint64 `pg:",use_zero"`
+	Hash         string `bun:",pk"`
+	Round        uint64
 	Header       *Header
-	Uncles       []*Header
-	Transactions []*Transaction `pg:"rel:has-many"`
+	Uncles       []*Header      `bun:"rel:has-many,join:hash=nephew_hash"`
```
ptrus, line 63: Is `Uncles` ever non-empty in oasis case? If it's always empty, we can remove the relation and just keep it as (an always empty) `[]*Header`

storage/psql/psql.go
```diff
@@ -2,54 +2,64 @@ package psql
...
 	"github.com/starfishlabs/oasis-evm-web3-gateway/model"
+
+	"github.com/uptrace/bun"
+	"github.com/uptrace/bun/dialect/pgdialect"
+	"github.com/uptrace/bun/driver/pgdriver"
 )
 
 type PostDB struct {
-	DB *pg.DB
+	DB *bun.DB
```
ptrus, line 22: https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/105 made some conflicts here.

The fixes needed are:
- change the PostDB struct to:
```golang
type PostDB struct {
	DB bun.IDB
}
```

And update `RunInTransaction` (added in #105) as follows:

```golang
func transactionStorage(t *bun.Tx) storage.Storage {
	db := PostDB{t}
	return &db
}

// RunInTransaction runs a function in a transaction. If function
// returns an error transaction is rolled back, otherwise transaction
// is committed.
func (db *PostDB) RunInTransaction(ctx context.Context, fn func(storage.Storage) error) error {
	bdb, ok := db.DB.(*bun.DB)
	if !ok {
		return fmt.Errorf("already in a transaction")
	}
	return bdb.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
		db := transactionStorage(&tx)
		return fn(db)
	})
}
```

Let me know if you'll need help

model/db.go
```diff
@@ -27,26 +27,26 @@ const Continues string = "continues"
...
 type IndexedRoundWithTip struct {
-	Tip   string `pg:",pk"`
+	Tip   string `bun:",pk"`
 	Round uint64
 }
 
 type AccessList []AccessTuple
 
 // Transaction is ethereum transaction.
 type Transaction struct {
```
ptrus, line 37: It's super simple to add additional indexes with `bun` so lets also address #96 here.

Example to add indexes on `block_hash` and `round` for  `Transaction`  table do:
```golang
var _ bun.AfterCreateTableHook = (*Transaction)(nil)

func (*Transaction) AfterCreateTable(ctx context.Context, query *bun.CreateTableQuery) error {
	_, err := query.DB().NewCreateIndex().
		Model((*Transaction)(nil)).
		Index("transaction_on_block_hash").
		Column("block_hash").
		Exec(ctx)
	if err != nil {
		return err
	}

	_, err = query.DB().NewCreateIndex().
		Model((*Transaction)(nil)).
		Index("transaction_on_round").
		Column("round").
		Exec(ctx)

	return err
}
```

Do similar for other tables as mentioned in the linked issue

SzeChuanChilliSauce: ok

storage/psql/psql.go
SzeChuanChilliSauce: ok, thanks.

model/db.go
SzeChuanChilliSauce: ok, i will remove the uncles of the block.

storage/psql/psql.go
anonymousGiga: If we use DB ```bun.IDB``` instead of ```DB *bun.DB```, these code not compile.    

```
	// PKs are required for ON CONFLICT DO UPDATE
		typ := reflect.TypeOf(value)
		table := db.DB.Table(typ)
		pks := make([]string, len(table.PKs))
		for i, f := range table.PKs {
			pks[i] = f.Name
		}
		_, err = db.DB.NewInsert().
			Model(value).
			On(fmt.Sprintf("CONFLICT (%v) DO UPDATE", strings.Join(pks, ","))).
			Exec(context.Background()) 
```

Can you give some advice?   @ptrus

anonymousGiga: We use this way(```bdb, ok := db.DB.(*bun.DB)```)  convert IDB to *bun.DB,  an error report in ```upsertSingle```.

ptrus: > We use this way(bdb, ok := db.DB.(*bun.DB)) convert IDB to *bun.DB, an error report in upsertSingle.

This is not ok, as in case `upsert` is called from a transaction it won't work (as it won't be able to cast since `IDB~ will be `*tx.Tx` )

ptrus (approved):
I see all tests are passing now, great!

Feel free to merge this, we can addres remaining comments in follow up PRs (config param for pool size, indexes, context changes on methods).

SzeChuanChilliSauce:
please review. @ptrus

anonymousGiga (approved):

ptrus:
>  we configure the pool size and add database indexes, context on methods.

Looks good, thanks!

PR #102, opened 2021-12-07
getLogs: limit number of rounds in single query

### LINKED ISSUE #99: Limit the range of logs that can be queried
Don't allow querying more than X (maybe 100?) rounds in a single getLogs query.

ptrus:
fixes: https://github.com/starfishlabs/oasis-evm-web3-gateway/issues/99

### DIFF
--- a/rpc/eth/api.go
+++ b/rpc/eth/api.go
@@ -23,6 +23,7 @@ import (
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/evm"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 
+	"github.com/starfishlabs/oasis-evm-web3-gateway/conf"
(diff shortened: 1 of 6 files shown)


rpc/eth/api.go
```diff
@@ -42,13 +42,19 @@ var (
...
 	// otherwise transactions without signature would be underestimated.
 	estimateGasSigSpec = estimateGasDummySigSpec()
 )
 
-const revertErrorPrefix = "reverted: "
+const (
+	revertErrorPrefix = "reverted: "
+
+	// Maximum number of rounds to query for in a get logs query.
+	getLogsMaxRounds = 100
```
kostko: Should we add this to config instead?

ptrus: yeah will move it there

ptrus: done

kostko (approved):

PR #104, opened 2021-12-08
fix bug cased by pr98

### LINKED ISSUE #103: Bugs caused by pr98
Problems caused by pr #98

anonymousGiga:
fix #103

### DIFF
--- a/indexer/backend.go
+++ b/indexer/backend.go
@@ -1,6 +1,7 @@
 package indexer
 
 import (
+	"database/sql"
 	"encoding/hex"
 	"errors"
 	"sync"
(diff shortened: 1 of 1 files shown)


SzeChuanChilliSauce (approved):

ptrus:
Hm how does this fix work? doesn't https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/f0392138655d2ed5bfe6120fecc1df9f7ecc02b9/storage/psql/psql.go#L164-L167 already catch the error? If not, can this fix be moved there?

ptrus:
Can you add a test for the error that this fixes, so that we have this tested

anonymousGiga:
OK，maybe we missed this place when we rebase from main. I'll deleted this.

PR #109, opened 2021-12-09
indexer: fix indexing round 0

ptrus:
Before, indexing always started at `QueryLastIndexedRound() + 1` (adjusted by node pruning).

This was incorrect for the case when no rounds are indexed, as in that case the `QueryLastIndexedRound()` returned `0` and indexing started at round `1`

This also caused `earliest` block queries to fail (when the earliest round was 0 and there was no pruning), as it tried to query round `0` which was not indexed.

### DIFF
--- a/indexer/backend.go
+++ b/indexer/backend.go
@@ -217,7 +217,7 @@ func (p *psqlBackend) storeIndexedRound(round uint64) error {
 
 // QueryLastIndexedRound returns the last indexed round.
 func (p *psqlBackend) QueryLastIndexedRound() (uint64, error) {
-	indexedRound, err := p.storage.GetContinuesIndexedRound()
+	indexedRound, err := p.storage.GetLastIndexedRound()
 	if err != nil {
 		return 0, err
 	}
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -118,19 +118,31 @@ func (s *Service) indexingWorker() {
 			continue
 		}
 
+		var startAt uint64
+		// Get last indexed round.
 		lastIndexed, err := s.backend.QueryLastIndexedRound()
-		if err != nil {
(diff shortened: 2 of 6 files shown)


indexer/indexer.go
```diff
@@ -118,19 +118,31 @@ func (s *Service) indexingWorker() {
...
 				"err", err,
 			)
 			continue
 		}
-		if latest < lastIndexed {
+		if latest < startAt {
 			panic("This is a new chain, please clear the db first!")
 		}
+		// Latest round already indexed.
+		if latest == lastIndexed {
```
kostko: Wouldn't this be problematic in case the error is `ErrNoRoundsIndexed` and both `lastIndexed` and `latest` are `0`?

indexer/backend.go
```diff
@@ -217,7 +217,7 @@ func (p *psqlBackend) storeIndexedRound(round uint64) error {
 
 // QueryLastIndexedRound returns the last indexed round.
 func (p *psqlBackend) QueryLastIndexedRound() (uint64, error) {
-	indexedRound, err := p.storage.GetContinuesIndexedRound()
+	indexedRound, err := p.storage.GetLastIndexedRound()
 	if err != nil {
 		return 0, err
```
kostko, line 222: Instead of zero this could return the invalid round (`roothash.RoundInvalid`) in case of an error.

ptrus: yeah maybe, but it might be a bit confusing since the same value is used as the "latest" round as well, right?

indexer/indexer.go
ptrus: Hm yeah, should move this into the "have rounds indexed" case.

kostko (approved):

PR #111, opened 2021-12-13
add truncate-db command

### LINKED ISSUE #97: Add subcommand to reindex the database
It is useful to be able to have a subcommand that would enable one to trigger a database reindex. This way people wouldn't need to drop and re-create the database in order to trigger reindexing.

ptrus:
Fixes: https://github.com/starfishlabs/oasis-evm-web3-gateway/issues/97

Adds a command to truncate the db.

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,11 +12,13 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/client"
 	"github.com/spf13/cobra"
+	"github.com/uptrace/bun"
 	"google.golang.org/grpc"
 
 	"github.com/starfishlabs/oasis-evm-web3-gateway/conf"
 	"github.com/starfishlabs/oasis-evm-web3-gateway/indexer"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -49,6 +65,40 @@ func exec(cmd *cobra.Command, args []string) error {
...
+		return fmt.Errorf("unsafe not allowed")
+	}
+
+	// Initialize db.
+	db, err := psql.InitDB(ctx, cfg.Database)
+	if err != nil {
+		logger.Error("failed to initialize db", "err", err)
+		return err
+	}
+	if err := model.TruncateModel(ctx, db.DB.(*bun.DB)); err != nil {
```
kostko, line 92: Does this also handle schema changes?

ptrus: Hm, no it doesn't. So the only case where it would leave tables is that if a table is removed from the model, it would be leftover. So one would need to manually delete those tables.

The alternative is to first do a raw query to get all table names from `pg_tables`, and then delete the tables. Maybe there should be another flag to the truncate model (`--drop-all`) to force this behavior.

ptrus:
This will be done in https://github.com/starfishlabs/oasis-evm-web3-gateway/pull/113

PR #112, opened 2021-12-13
implement /health endpoint

### LINKED ISSUE #86: Health check endpoint
There should be a health check endpoint that can be reached via an HTTP GET request. It should report the health status of the gateway by performing some standard queries against the latest round (could be done periodically in a background goroutine so that health check requests don't cause too much load).

This would help for things like load balancers which currently rely just on TCP connection establishment to know whether the gateway is available.

ptrus:
Fixes: https://github.com/starfishlabs/oasis-evm-web3-gateway/issues/86

Implements `/health` endpoint.

The health-check is:
- query db for latest indexed round
- query node for latest round
- compare rounds (ensure these are not too far apart)

Note: oasis-node status is not queried (as was mentioned in the issue) so that gateway does not need access to the `Control` endpoint of the oasis-node.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -13,6 +13,14 @@ import (
 	"github.com/starfishlabs/oasis-evm-web3-gateway/storage"
 )
 
+// HealthCheck is the health checker interface.
+type HealthCheck interface {
+	// Health checks the health of the service.
+	//
+	// This method should return quickly and should avoid performing any blocking operations.
+	Health() error
+}
+
 type Server struct {
 	Config *conf.Config
 	Web3   *Web3Gateway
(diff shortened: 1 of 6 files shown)


server/server.go
```diff
@@ -13,6 +13,12 @@ import (
 	"github.com/starfishlabs/oasis-evm-web3-gateway/storage"
 )
 
+// HealthCheck is the health checker interface.
+type HealthCheck interface {
+	// Health checks the health of the service.
```
kostko, line 18: Maybe add a comment to say that this should not perform any blocking operations and should return quickly.

kostko (approved):

server/server.go
ptrus: good idea :+1:

PR #113, opened 2021-12-13
db: add support for migrations

### LINKED ISSUE #97: Add subcommand to reindex the database

ptrus:
- Sets up database migrations and adds the initial migration
- adds truncate database command (fixes: #97) 
- removes unneeded DB models (`BlockRefs`, `TransactionRefs`)
  - `TransactionRefs` was completely unused, `BlockRefs` contained duplicate fields already present in `Blocks`

### DIFF
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -15,8 +15,7 @@ import (
 )
 
 const (
-	storageRequestTimeout            = 5 * time.Second
-	storageRetryTimeout              = 1 * time.Second
+	indexerLoopDelay                 = 1 * time.Second
 	pruningCheckInterval             = 60 * time.Second
 	healthCheckInterval              = 10 * time.Second
 	healthCheckIndexerDriftThreshold = 10
@@ -169,13 +168,12 @@ func (s *Service) indexingWorker() {
 		select {
 		case <-s.ctx.Done():
 			return
-		default:
+		case <-time.After(indexerLoopDelay):
 		}
 
 		// Query latest round available at the node.
 		latest, err := s.getRoundLatest()
 		if err != nil {
-			time.Sleep(storageRequestTimeout)
 			s.Logger.Info("failed to query latest round",
 				"err", err,
 			)
@@ -200,7 +198,6 @@ func (s *Service) indexingWorker() {
 			}
 			// Latest round already indexed.
 			if latest == lastIndexed {
-				time.Sleep(storageRetryTimeout)
 				continue
 			}
 			startAt = lastIndexed + 1
@@ -209,7 +206,6 @@ func (s *Service) indexingWorker() {
 		// Get last retained round on the node.
 		lastRetainedBlock, err := s.client.GetLastRetainedBlock(s.ctx)
 		if err != nil {
-			time.Sleep(storageRequestTimeout)
 			s.Logger.Error("failed to retrieve last retained round",
 				"err", err,
 			)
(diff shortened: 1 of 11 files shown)


model/init.go
```diff
@@ -2,41 +2,75 @@ package model
...
-	for _, tb := range tables {
-		if _, err := db.NewDropTable().Model(tb).IfExists().Exec(ctx); err != nil {
-			return err
-		}
+// Migrate migrates the DB to latest version.
+func Migrate(ctx context.Context, db *bun.DB) error {
+	logger := logging.GetLogger("migration")
+
+	// Initialize the migrator.
+	migrator := migrate.NewMigrator(db, migrations.Migrations)
```
ptrus, line 60: Hm actually it would be better to embed these migrations in the binary so that one only needs the binary to do the migrations.

Will look into it.

kostko (commented):
Is there a way to define migrations as Go code?

ptrus:
> Is there a way to define migrations as Go code?

Yeah, it is, this is how it looks like:

```golang
func init() {
	Migrations.MustRegister(func(ctx context.Context, db *bun.DB) error {
		fmt.Print(" [up migration] ")
		return nil
	}, func(ctx context.Context, db *bun.DB) error {
		fmt.Print(" [down migration] ")
		return nil
	})
}
```

but for this initial migration we would either need to duplicate all the Go model definition structs (since changing the model in code should not change the migration, instead a new migration should be done for the changes) or make raw queries. Therefore, using SQL migration for this initial one made sense to me (also could generate the SQL by dumping the existing db schema).

I think the Go migrations will make most sense in any future migrations that will also modify existing DB data, as that will be much nicer to code in Go.

kostko:
I see, yeah that makes sense for the initial migration. Then maybe use go:embed to avoid the need to ship the files separately.

ptrus:
> I see, yeah that makes sense for the initial migration. Then maybe use go:embed to avoid the need to ship the files separately.

Done. Also migrations are now run automatically on gateway start, so no need to issue the separate migration command.

model/migrations/initial.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/uptrace/bun"
+	"github.com/uptrace/bun/migrate"
+)
+
+// Migrations are all migrations.
+var Migrations *migrate.Migrations
+
+//go:embed 20211213143752_initial.up.sql
+var initialUpSQL []byte
```
kostko: It seems that [this could be of type `embed.FS`](https://pkg.go.dev/embed) and then you would be able to just use the upstream `NewSQLMigrationFunc`.

ptrus: oh nice

indexer/indexer.go
```diff
@@ -15,8 +15,7 @@ import (
 )
 
 const (
-	storageRequestTimeout            = 5 * time.Second
-	storageRetryTimeout              = 1 * time.Second
+	indexerLoopTimeout               = 1 * time.Second
```
kostko: Maybe call it `indexerLoopDelay`.

model/init.go
```diff
@@ -2,41 +2,79 @@ package model
...
-	new(Transaction),
-	new(TransactionRef),
-	new(IndexedRoundWithTip),
-	new(Receipt),
-	new(Log),
-}
+// TruncateDatabase deletes all database tables in the `public` schema of the configured database.
+//
+// Note: this method assumes that PostgresSQL is used as the underlying db.
+func TruncateDatabase(ctx context.Context, db *bun.DB) error {
```
kostko: Maybe rename this method to something like `DropTables`.

kostko (approved):

PR #115, opened 2021-12-14
fix issue 41,42,60

### LINKED ISSUE #41: nit:  not using defer
Stylistic inconsistency, since `defer` is used on L113 for `srv.startStopLock.Unlock()`

https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/f6a1b3ce1622603f9f2226da63a6644a2891802c/server/server.go#L111-L138

L118, L121, and L130 all do `srv.lock.Unlock()` which could have been a `defer srv.lock.Unlock()` after L115.  The only potential reason I can think of why it couldn't be written in the more idiomatic fashion would be if L134's call to `srv.doClose(nil)` had to be called without holding the lock, but inspection of that code

https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/f6a1b3ce1622603f9f2226da63a6644a2891802c/server/server.go#L165-L178

does not show any such requirement.  (If there _is_, then a comment would be appropriate.)  That the code before and after L130 both check `err != nil` is weird -- but from the git logs it appears the code was this way from when it was first added.

### LINKED ISSUE #42: code smell: errs is the nil slice -- comment is wrong?
https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/f6a1b3ce1622603f9f2226da63a6644a2891802c/server/server.go#L153-L156

There are no resources acquired by `Start()` and `errs` is never assigned to, so will have the default `nil` value.  This makes the `initializingState` and `runningState` handling the same.  This code is weird -- feels perhaps unfinished? or there was an intent to have `Start()` acquire resources at some point?

There is the channel `srv.stop` initialized in `New`, so maybe what this is referring to is to ensure goroutines will be in a clean state?  Except the `Server.Wait()` which calls `Web3.Wait()` appear to be the only place where `<-srv.stop` is done, and there are no other references to this channel, so this is soley for `runRoot` to block until the server is down, which is fine, but it's not really a resource cleanup but rather a synchronization task so cobra doesn't get confused.

### LINKED ISSUE #60: nit: API has error return; error is ignored
A `_` is used to ignore the error return from `IORoot.MarshalBinary()` at

https://github.com/starfishlabs/oasis-evm-web3-gateway/blob/22eb39c828e9319dbbb9e14f9fe1be9192f8715f/rpc/eth/api.go#L109

while the implementation

https://github.com/oasisprotocol/oasis-core/blob/94237e2be3121d47b6997780122f95f6a0d00756/go/common/crypto/hash/hash.go#L37-L40

does not currently return a non-nil error, the API allows it, and future iterations may.  While this is understandbly going to be difficult to test (need to mock oasis-core/.../hash) and the error handling is effectively dead code, it is a good idea to do something, e.g., panic or otherwise detect if/when a future change in the implementation introduces an error return. Such an update would not be a "breaking" API change and the major version number in the semantic version would not change, and generally the expectation is that dependent codebases will continue to work -- or, in this case, fail gracefully and allow easy debugging/upgrading via a nice panic message.

anonymousGiga:
fix issue #41 #42 #60

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -120,12 +120,11 @@ func (srv *Web3Gateway) Start() error {
 	defer srv.startStopLock.Unlock()
 
 	srv.lock.Lock()
+	defer srv.lock.Unlock()
 	switch srv.state {
 	case runningState:
-		srv.lock.Unlock()
 		return ErrServerRunning
 	case closedState:
-		srv.lock.Unlock()
 		return ErrServerStopped
 	}
 	srv.state = runningState
@@ -134,7 +133,6 @@ func (srv *Web3Gateway) Start() error {
 	if err != nil {
 		srv.stopRPC()
 	}
-	srv.lock.Unlock()
 
 	// Check if RPC endpoint startup failed.
 	if err != nil {
@@ -159,8 +157,7 @@ func (srv *Web3Gateway) Close() error {
 		return srv.doClose(nil)
 	case runningState:
 		// The server was started, release resources acquired by Start().
-		var errs []error
-		return srv.doClose(errs)
+		return srv.doClose([]error{errors.New("error in function close")})
 	case closedState:
 		return ErrServerStopped
 	default:


anonymousGiga:
@ptrus Please review, thanks.

server/server.go
```diff
@@ -160,6 +158,7 @@ func (srv *Web3Gateway) Close() error {
 	case runningState:
 		// The server was started, release resources acquired by Start().
 		var errs []error
+		errs = append(errs, errors.New("error in function close"))
```
ptrus:
```suggestion
		return srv.doClose([]error{errors.New("error in function close")})
```

ptrus (approved):

PR #116, opened 2021-12-14
indexer: minor fixes

ptrus:
- ore-allocate slices where sizes are known
- avoid calling `eth2DbLogs` twice
- fix `dbLogs` to contain all block logs (before it contained only last transaction logs (this is only used in the event system))

### DIFF
--- a/indexer/utils.go
+++ b/indexer/utils.go
@@ -94,15 +94,15 @@ func convertToEthFormat(
 		BaseFee:     baseFee.String(),
 	}
 
-	innerTxs := []*model.Transaction{}
-	innerReceipts := []*model.Receipt{}
+	innerTxs := make([]*model.Transaction, 0, len(transactions))
+	innerReceipts := make([]*model.Receipt, 0, len(transactions))
 	cumulativeGasUsed := uint64(0)
 	for idx, ethTx := range transactions {
 		v, r, s := ethTx.RawSignatureValues()
 		signer := ethtypes.LatestSignerForChainID(ethTx.ChainId())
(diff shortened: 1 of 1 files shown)


indexer/utils.go
```diff
@@ -94,16 +94,16 @@ func convertToEthFormat(
 		BaseFee:     baseFee.String(),
 	}
 
-	innerTxs := []*model.Transaction{}
-	innerReceipts := []*model.Receipt{}
+	innerTxs := make([]*model.Transaction, len(transactions))
+	innerReceipts := make([]*model.Receipt, len(transactions))
```
kostko: You could also do something like `make([]*model.Transaction, 0, len(transaction))` and leave the other parts as-is. This will allocate a slice with the given capacity and zero length.

ptrus: Yeah makes sense

indexer/utils.go
```diff
@@ -299,15 +299,15 @@ func (ib *indexBackend) StoreBlockData(oasisBlock *block.Block, txResults []*cli
...
 // eth2DbLogs converts ethereum log to model log.
 func eth2DbLogs(ethLogs []*ethtypes.Log) []*model.Log {
-	res := []*model.Log{}
+	res := make([]*model.Log, 0, len(ethLogs))
 
 	for _, log := range ethLogs {
-		topics := []string{}
-		for _, tp := range log.Topics {
-			topics = append(topics, tp.Hex())
+		topics := make([]string, len(log.Topics))
```
kostko: Also here (to be consistent).

kostko (approved):

PR #133, opened 2021-12-24
fix issue131

### LINKED ISSUE #131: `getBlock`: `extraData` should return `"0x" instead of `""`
Apparently this is what is expected in the response. (todo: confirm other gateways, e.g. infura)

anonymousGiga:
fix #131

### DIFF
--- a/rpc/utils/utils.go
+++ b/rpc/utils/utils.go
@@ -107,6 +107,10 @@ func ConvertToEthBlock(block *model.Block, fullTx bool) map[string]interface{} {
 		"totalDifficulty":  (*hexutil.Big)(big.NewInt(0)),
 	}
 
+	if len(header.Extra) == 0 {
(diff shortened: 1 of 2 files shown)


rpc/utils/utils.go
```diff
@@ -107,6 +107,10 @@ func ConvertToEthBlock(block *model.Block, fullTx bool) map[string]interface{} {
 		"totalDifficulty":  (*hexutil.Big)(big.NewInt(0)),
 	}
 
+	if len(header.Extra) == 0 {
```
ptrus, line 110: Nice, can you also add a simple test case for this?

anonymousGiga: Done.

ptrus: Thanks!

ptrus (approved):

PR #147, opened 2022-01-11
feat: automate releaes using goreleaser

### LINKED ISSUE #93: Set up versioning and release process
Once the codebase stabilizes we need to set up a versioning scheme and release process.

ptrus:
Fixes: https://github.com/oasisprotocol/emerald-web3-gateway/issues/93

- goreleaser is used to build the binaries
- changelog is generated from commits

TODO:
- [x] add a release process document
- [ ] after the initial release, start using commit messages to group change log sections (see `changelog.groups` in goreleaser.yml)

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,65 @@
+project_name: Emerald Web3 Gateway
+
+before:
+  hooks:
+    - go mod tidy
+
+builds:
+  - binary: emerald-web3-gateway
+    env:
+      - CGO_ENABLED=0
+    flags:
+      - -trimpath
+    ldflags:
+      - -buildid=
+      - "{{.Env.GOLDFLAGS_VERSION}}"
+    goos:
+      - linux
+    goarch:
+      - amd64
+
+archives:
+  - name_template: "{{replace .ProjectName \" \" \"_\" | tolower}}_{{.Version}}_{{.Os}}_{{.Arch}}"
+    wrap_in_directory: true
+
+checksum:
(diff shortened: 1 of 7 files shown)


.goreleaser.yaml
```diff
@@ -0,0 +1,65 @@
...
+changelog:
+  sort: asc
+  use: github
+  filters:
+    exclude:
+    - Merge pull request
+    - Merge remote-tracking branch
+    - Merge branch
+
+  # Enable this after the initial release.
```
ptrus, line 41: The initial release will be done without these (so all commits will be in the same section), afterwards the commit messages should follow the below format.

ptrus: Should a gitlint rule could be added to ensure all commits are in the format of:

```
breaks[\w]*:|fix[(\w)]*:|feat[(\w)]*:|other[(\w)]*:|deps[(\w)]*:
```

kostko (commented):
Is the plan to run this automatically when specific tags are pushed?

ptrus:
Yeah this is setup the way that it is run automatically when a tag is pushed, see: https://github.com/oasisprotocol/emerald-web3-gateway/blob/main/docs/release-process.md#release-process  The only manual part is pushing the appropriate tag. (this is similar process as the goreleaser process itself uses: https://github.com/goreleaser/goreleaser)

ptrus:
Oh i see, the github action is missing in the PR :-) let me push it

kostko (approved):

PR #168, opened 2022-01-31
Add Docker support

### LINKED ISSUE #167: Create Docker image for local dApp development
Create a docker image containing a complete stack for local dApp development on Emerald:
- oasis-node
- emerald-paratime
- oasis-net-runner with configured fixtures for emerald testing accounts
- postgres
- emerald-web3-gateway listening on 8545

Some open questions and considerations:
- could we use the code from ci-test.yaml to prepare the image?
- `geth --dev` spins up random testing accounts each time. Could we achieve the same? Could we offer folks also using Oasis CLI to perform deposits/withdrawals?
- when should we bump this image and how to version it? On new oasis-core release? Emerald-paratime release? Emerald-web3-gateway release? oasis-sdk release (in case of using Oasis CLI)? All of those?

matevz:
Fixes #167:
- adds Dockerfile for creating local dApp development image

Demo:
```
$ docker run -it -p8545:8545 -p8546:8546 emerald-dev
emerald-dev v22.1 (oasis-core: v21.3.9, emerald-paratime: v6.2.0, emerald-web3-gateway: v1.1.0-git4d6203a+dirty)

Starting oasis-net-runner with Emerald ParaTime...
Starting postgresql...
Starting emerald-web3-gateway...
Populating account(s) (this might take a moment)...

Available Accounts
==================
(0) 0xB1c9D56d0292912335BE28245ab90B4E6f8E71A8 (100 ROSE)
(1) 0x0207d6F23a714cf491b0274b5F2faA5DFb6F907C (100 ROSE)
(2) 0xCC6BEba2F452ADcfB6E625A91A830472db6c0681 (100 ROSE)
(3) 0x47BF64695DBDF74880D9ca9AdB0e5563ee3cBf01 (100 ROSE)
(4) 0xd84C6529926c87e46F59B065ACeE31006eB03886 (100 ROSE)

Private Keys
==================
(0) 0x34635f9fb1c5915088770c7b49c962eebc7e305e784bbe332d7f0f8ca8d94af4
(1) 0xe195da3f70a53760a16959ef4658b686c2f61ce2ce043f8ade3110d1c50d7814
(2) 0xf6aefdee319f83ec33c561294c8d3b465cf3143c9350f0e47603140f49c28b20
(3) 0x7cf1586f6f5a9405143e5dde3a848030c92fdd4e6fa04d9e41b25a55562d23fc
(4) 0x11f63e8a6697afcfd071020dfecf13bd3aab2b214d30c93ef6664b71a7658e26

HD Wallet
==================
Mnemonic:       pink describe dose survey primary warfare rigid pond want type document chunk
Base HD Path:   m/44'/60'/0'/0/%d

WARNING: Emerald is running in ephemeral mode. The chain state will be lost after you quit!

Listening on http://localhost:8545 and ws://localhost:8546
```


conf/docker.yml
```diff
@@ -0,0 +1,30 @@
+runtime_id: "8000000000000000000000000000000000000000000000000000000000000000"
```
ptrus: If you want to avoid this file you could also set all settings via ENV vars.

docker/emerald-dev/README.md
```diff
@@ -0,0 +1,23 @@
...
+set up your local development environment for dApps running on Emerald:
+- oasis-node and oasis-net-runner
+- emerald paratime
+- emerald-web3-gateway
+
+To build the docker image, run the following in your emerald-web3-gateway root
+folder:
+
+```sh
+docker build -t emerald-dev -f docker/emerald-dev/Dockerfile .
```
ptrus: Could also build the image in CI (on every merge to main) and push it to dockerhub

conf/docker.yml
matevz: This was now moved to `docker/` so it should be ok.

docker/emerald-dev/README.md
matevz: Also for new oasis-core, emerald-paratime releases, which might get tricky because of the hooks. Let's deploy the image manually for now. Opened a separate issue for this https://github.com/oasisprotocol/emerald-web3-gateway/issues/169

docker/emerald-dev/Dockerfile
```diff
@@ -0,0 +1,59 @@
+FROM golang:1.17 AS emerald-web3-gateway
+
+ARG OASIS_EMERALD_WEB3_GATEWAY_TAG=matevz/docker
+
+#RUN git clone -b ${OASIS_EMERALD_WEB3_GATEWAY_TAG} https://github.com/oasisprotocol/emerald-web3-gateway.git
```
ptrus: This should probably be uncommented or removed?

docker/emerald-dev/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+COPY . /go/emerald-web3-gateway
+RUN cd emerald-web3-gateway && go build
```
ptrus: There is a Makefile in this repository, could this use it instead of running `go build` manually?

Also, could add a target for building the `oasis-deposit` to the Makefile as well.

docker/emerald-dev/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+RUN cd emerald-web3-gateway/docker/emerald-dev/oasis-deposit && go build
+
+FROM ubuntu:20.04
+
+ARG OASIS_CORE_VERSION=21.3.9
+ENV OASIS_EMERALD_VERSION=6.2.0
```
ptrus: Does this need to be an `ENV`? Is it used for anything other than building the image?

docker/emerald-dev/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+
+# emerald-web3-gateway binary, config, spinup-* scripts and staking_genesis.json.
+COPY --from=emerald-web3-gateway /go/emerald-web3-gateway/emerald-web3-gateway ${OASIS_EMERALD_WEB3_GATEWAY}
+COPY --from=emerald-web3-gateway /go/emerald-web3-gateway/docker/emerald-dev/oasis-deposit/oasis-deposit ${OASIS_DEPOSIT}
+COPY docker/emerald-dev/emerald-dev.yml ${OASIS_EMERALD_WEB3_GATEWAY_CONFIG_FILE}
+COPY docker/emerald-dev/start-emerald-dev.sh /
+COPY tests/tools/* /
+
+# Configure oasis-node and oasis-net-runner.
+RUN sed -E -i 's#http://[^\s]*archive\.ubuntu\.com/ubuntu#http://si.archive.ubuntu.com/ubuntu#g' /etc/apt/sources.list
```
ptrus: Why force `si.archive.ubuntu.com`  :grin:

docker/emerald-dev/README.md
```diff
@@ -0,0 +1,24 @@
...
+- oasis-node and oasis-net-runner
+- emerald paratime
+- emerald-web3-gateway
+- oasis-deposit helper to fund your ethereum wallet
+
+To build the docker image, run the following in your emerald-web3-gateway root
+folder:
+
+```sh
+docker build -t emerald-dev -f docker/emerald-dev/Dockerfile .
```
ptrus: Maybe add a `make docker` (or `make build-docker`) target to the repository Makefile

docker/emerald-dev/README.md
```diff
@@ -0,0 +1,24 @@
...
+```
+
+To run the Emerald development environment locally run:
+
+```sh
+docker run -it -p8545:8545 -p8546:8546 emerald-dev
+```
+
+You can now connect with your hardhat, truffle or metamask to
```
ptrus, line 23: Maybe a note that state is lost (reset) if the container is restarted would be nice in the readme as well.

ptrus (commented):
Looks good, some minor comments

docker/emerald-dev/Dockerfile
gw0: I believe the ideas was to use either this or the `COPY` command below... Instead of `COPY` it would be more efficient to allow mounting a host volume (e.g. `-v /home/.../emerald/web3-gateway:/go/emerald-web3-gateway`).

It could even be conditional (if something is mounted do nothing, else download from Git)?
```
RUN [ ! -d "/go/emerald-web3-gateway"] \
 && git clone -b ${OASIS_EMERALD_WEB3_GATEWAY_TAG} https://github.com/oasisprotocol/emerald-web3-gateway.git
```

docker/emerald-dev/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+ADD "https://github.com/oasisprotocol/oasis-core/releases/download/v${OASIS_CORE_VERSION}/oasis_core_${OASIS_CORE_VERSION}_linux_amd64.tar.gz" /
+RUN tar xfvz "oasis_core_${OASIS_CORE_VERSION}_linux_amd64.tar.gz" && \
```
gw0: Why `&&` on the end of the line? In Dockerfiles in-the-wild it is more common to put in in the beginning of the next line with a space in front (such that it aligns with `RUN`). Otherwise, the new trend is to first do `set -eux` (to fail on errors) and then use `; \`.

docker/emerald-dev/Dockerfile
gw0: It seems to be used in `tests/tools/spinup-oasis-stack.sh`?

docker/emerald-dev/README.md
gw0: Just to let you know: If you run this in the root directory of a repository it will send all the files it can find to the Docker context. If there are many files, this can take a while (especially, if the Docker daemon is running on another machine).

docker/emerald-dev/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+# emerald-web3-gateway binary, config, spinup-* scripts and staking_genesis.json.
+COPY --from=emerald-web3-gateway /go/emerald-web3-gateway/emerald-web3-gateway ${OASIS_EMERALD_WEB3_GATEWAY}
+COPY --from=emerald-web3-gateway /go/emerald-web3-gateway/docker/emerald-dev/oasis-deposit/oasis-deposit ${OASIS_DEPOSIT}
+COPY docker/emerald-dev/emerald-dev.yml ${OASIS_EMERALD_WEB3_GATEWAY_CONFIG_FILE}
+COPY docker/emerald-dev/start-emerald-dev.sh /
+COPY tests/tools/* /
+
+# Configure oasis-node and oasis-net-runner.
+RUN sed -E -i 's#http://[^\s]*archive\.ubuntu\.com/ubuntu#http://si.archive.ubuntu.com/ubuntu#g' /etc/apt/sources.list
+ADD "https://github.com/oasisprotocol/oasis-core/releases/download/v${OASIS_CORE_VERSION}/oasis_core_${OASIS_CORE_VERSION}_linux_amd64.tar.gz" /
```
gw0, line 32: If you want to minimize the number of Docker image layers, you could replace this with `curl` or `wget` and combined with other related commands in one `RUN`.

docker/emerald-dev/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+	mkdir -p "$(dirname ${OASIS_NODE})" "$(dirname ${OASIS_NET_RUNNER})" && \
+	mv "oasis_core_${OASIS_CORE_VERSION}_linux_amd64/oasis-node" "${OASIS_NODE}" && \
+	mv "oasis_core_${OASIS_CORE_VERSION}_linux_amd64/oasis-net-runner" "${OASIS_NET_RUNNER}"
+
+# Configure Emerald ParaTime.
+RUN mkdir -p "$(dirname ${OASIS_EMERALD_PARATIME})"
+ADD "https://github.com/oasisprotocol/emerald-paratime/releases/download/v${OASIS_EMERALD_VERSION}/emerald-paratime" "${OASIS_EMERALD_PARATIME}"
+RUN chmod a+x "${OASIS_EMERALD_PARATIME}"
+
+# Configure Bubblewrap and Postgresql packaged by Ubuntu.
```
gw0: Hm, is Bubblewrap still needed (due to `runtime.provisioner=unconfined`)?

gw0 (commented):
A couple more suggestions.

docker/emerald-dev/start-emerald-dev.sh
```diff
@@ -0,0 +1,47 @@
...
+EMERALD_WEB3_GATEWAY_PID=$!
+
+echo "Populating ethereum account..."
+# Wait for compute nodes before initiating deposit.
+${OASIS_NODE} debug control wait-ready -a unix:${OASIS_NODE_SOCKET}
+
+${OASIS_DEPOSIT} -sock unix:${OASIS_NODE_SOCKET} -to ${ETHEREUM_ADDRESS} >/dev/null
+
+echo "Test account address: ${ETHEREUM_ADDRESS}"
+echo "Test account private key: ${ETHEREUM_PRIVATE_KEY}"
```
peterjgilbert: Can you also print the mnemonic?

docker/emerald-dev/Dockerfile
matevz: curl and wget are not installed in the base ubuntu image so I decided to just use the docker's ADD instead.

docker/emerald-dev/Dockerfile
matevz: Moved `&&`s at the beginning of new lines, nice.

I didn't manage to get `;` working in Dockerfile though :thinking: I always use `set -eux` in the scripts, but here each `RUN` is probably executed in its own shell and settings like these get forgotten on the next run.

docker/emerald-dev/Dockerfile
matevz: I didn't find `-v` for `docker build`. Are you sure this is not just a `docker run` functionality?

docker/emerald-dev/README.md
matevz: Yeah, this is fine. At least for the near-term future.

docker/emerald-dev/Dockerfile
gw0: Oops, you are right, this can not work this way.

docker/emerald-dev/start-emerald-dev.sh
```diff
@@ -0,0 +1,56 @@
...
+/etc/init.d/postgresql start >/dev/null
+
+echo "Starting emerald-web3-gateway..."
+# Wait for oasis-node socket before starting web3 gateway.
+while ! [[ -S ${OASIS_NODE_SOCKET} ]]; do sleep 1; done
+
+${EMERALD_WEB3_GATEWAY} --config ${EMERALD_WEB3_GATEWAY_CONFIG_FILE} 2>1&>/var/log/emerald-web3-gateway.log &
+EMERALD_WEB3_GATEWAY_PID=$!
+
+echo "Populating ethereum account(s)..."
```
ptrus: Maybe also echo something like `"(this might take a minute)"` here, as this step does take some time (due to waiting for initial runtime rounds to be processed).

ptrus (approved):
Looks good to me!

PR #171, opened 2022-02-06
EIP-1898 implementation

badangrycloud:


rpc/eth/api.go
```diff
@@ -25,9 +25,9 @@ import (
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/evm"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
 
-	"github.com/oasisprotocol/emerald-web3-gateway/conf"
-	"github.com/oasisprotocol/emerald-web3-gateway/indexer"
-	"github.com/oasisprotocol/emerald-web3-gateway/rpc/utils"
+	"github.com/oasisprotocol/oasis-evm-web3-gateway/conf"
```
ptrus: Please revert these import renames

rpc/eth/api.go
```diff
@@ -167,11 +167,11 @@ func (api *PublicAPI) GetBlockTransactionCountByNumber(blockNum ethrpc.BlockNumb
 	return hexutil.Uint(n), nil
 }
 
-func (api *PublicAPI) GetStorageAt(address common.Address, position hexutil.Big, blockNum ethrpc.BlockNumber) (hexutil.Big, error) {
-	logger := api.Logger.With("method", "eth_getStorageAt", "address", address, "position", position, "block_number", blockNum)
+func (api *PublicAPI) GetStorageAt(address common.Address, position hexutil.Big, blockNrOrHash ethrpc.BlockNumberOrHash) (hexutil.Big, error) {
+	logger := api.Logger.With("method", "eth_getStorageAt", "address", address, "position", position, "block number or hash", blockNrOrHash)
```
ptrus:
```suggestion
	logger := api.Logger.With("method", "eth_getStorageAt", "address", address, "position", position, "block_or_hash", blockNrOrHash)
```

Same in other places.

rpc/eth/api.go
```diff
@@ -673,3 +674,19 @@ func (api *PublicAPI) Hashrate() hexutil.Uint64 {
...
 	return 0
 }
+
+// getBlockRound returns the block round from BlockNumberOrHash.
+func (api *PublicAPI) getBlockRound(logger *logging.Logger, blockNrOrHash ethrpc.BlockNumberOrHash) (uint64, error) {
+	switch {
+	case blockNrOrHash.BlockHash == nil && blockNrOrHash.BlockNumber == nil:
+		return 0, fmt.Errorf("types BlockHash and BlockNumber cannot be both nil")
+	case blockNrOrHash.BlockHash != nil:
+		round, err := api.backend.QueryBlockRound(*blockNrOrHash.BlockHash)
```
ptrus:
```suggestion
		return api.backend.QueryBlockRound(*blockNrOrHash.BlockHash)
```

rpc/eth/api.go
```diff
@@ -673,3 +674,19 @@ func (api *PublicAPI) Hashrate() hexutil.Uint64 {
 	logger.Debug("request")
 	return 0
 }
+
+// getBlockRound returns the block round from BlockNumberOrHash.
+func (api *PublicAPI) getBlockRound(logger *logging.Logger, blockNrOrHash ethrpc.BlockNumberOrHash) (uint64, error) {
+	switch {
+	case blockNrOrHash.BlockHash == nil && blockNrOrHash.BlockNumber == nil:
+		return 0, fmt.Errorf("types BlockHash and BlockNumber cannot be both nil")
+	case blockNrOrHash.BlockHash != nil:
```
ptrus: According to the spec an error should be returned in case both `blockHash` and `blockNumber` are specified https://github.com/ethereum/EIPs/issues/2247#issuecomment-523167874

However, `ethrpc.BlockNumberOrHash` already handles that case when unmarshaling (https://github.com/ethereum/go-ethereum/blob/6ce4670bc03602bb6638c490f892e848647c15c0/rpc/types.go#L134-L136), so a separate check is not needed here at the moment. Maybe just add a comment that the case is already handled by `BlockNumberOrHash`.

rpc/eth/api.go
```diff
@@ -673,3 +674,19 @@ func (api *PublicAPI) Hashrate() hexutil.Uint64 {
...
+		round, err := api.backend.QueryBlockRound(*blockNrOrHash.BlockHash)
+		return round, err
+	case blockNrOrHash.BlockNumber != nil:
+		round, err := api.roundParamFromBlockNum(logger, *blockNrOrHash.BlockNumber)
```
ptrus:
```suggestion
		return api.roundParamFromBlockNum(logger, *blockNrOrHash.BlockNumber)
```

rpc/eth/api.go
```diff
@@ -599,7 +600,7 @@ func (api *PublicAPI) GetLogs(filter filters.FilterCriteria) ([]*ethtypes.Log, e
 	filtered := make([]*ethtypes.Log, 0, len(ethLogs))
 	for _, log := range ethLogs {
 		// Filter by address.
-		addressMatch := len(filter.Addresses) == 0
+		addressMatch := len(filtered) == 0
```
ptrus: Revert this change

rpc/eth/api.go
```diff
@@ -673,3 +674,19 @@ func (api *PublicAPI) Hashrate() hexutil.Uint64 {
 	logger.Debug("request")
 	return 0
 }
+
+// getBlockRound returns the block round from BlockNumberOrHash.
+func (api *PublicAPI) getBlockRound(logger *logging.Logger, blockNrOrHash ethrpc.BlockNumberOrHash) (uint64, error) {
```
ptrus: This function doesn't use the `logger *logging.Logger` so remove the argument.

ptrus (commented):
Thanks, looks mostly good, please address the comments.

rpc/eth/api.go
badangrycloud: this is use in the roundParamFromBlockNum func so is needed to that function it's called inside

ptrus (approved):
Looks good now. 

Two more things before this can get merged:
- squash commits in a single commit titled: `feat: implement EIP-1898`
- rebase on `main` branch
- 
Thanks!

PR #174, opened 2022-02-09
fix: log index should be position in the block

### LINKED ISSUE #172: Bug: eth_getLogs index
It seems that the log index returned is based on its position in the tx, not its position in the block as specified by `eth_getLogs`.
See: https://www.quicknode.com/docs/ethereum/eth_getLogs, “logindex - Integer of log index position in the block encoded as a hexadecimal. null if pending.”

ptrus:
Fixes: #172 

Also refactored the model/migrations structure a bit.

Had to also implement a custom migrator (based on the upstream one) as explained in the code:
```
// Package migrator implements a custom migrator inspired by the upstram [migrator](https://github.com/uptrace/bun/blob/master/migrate/migrator.go).
//
// The upstream migrator first marks the migration as applied and then only after runs it.
// This migrator runs both the migration and the "successful mark" in a transaction. Therefore if
// the migration is aborted (or fails) the transaction is not applied and no manual rollback is needed.
```
The upstream migrator has this weird behavior where it always first marks the transaction as applied and only after runs it (not even in a transaction). Meaning that in case of failures, one needs to manually rollback and re-apply the migration.

TODO:
- [x] consider writing a migration for this to avoid reindexing existing deployments
- [x] test migration on testnet state (~10 minutes to complete it)
- [x] test migration on miannet state (~15 minutes)


kostko (commented):
Looks good, would be great if a migration was possible to avoid reindex.

main.go
```diff
@@ -54,6 +54,13 @@ var (
 		Short: "migrate-db",
 		RunE:  migrateExec,
 	}
+
+	// Migrations unlock.
+	unlockCmd = &cobra.Command{
+		Use:   "unlock-migrations",
```
kostko: What does this do, when is it needed?

kostko (approved):

main.go
ptrus: So the upstream migrator uses a lock table to prevent multiple migrations running at the same time. The use case of the upstream migrator seems to be manually running migrations, so having a separate lock makes sense I guess to prevent starting multiple migrations at the same time - specially since they also don't use transactions. The table can stay locked if the process is killed while the migration is in progress.

However, with the custom migrator that uses transactions and our use case where the app runs the migrations automatically, I think the lock table isn't actually needed anymore, so I think I'll just remove it altogether.

ptrus: Fixed, removed the separate lock table. Also added a test that tries to concurrently migrate and ensures the migrations are run only once.

PR #191, opened 2022-02-25
indexer: query max block gas limit

### LINKED ISSUE #151: Hardcoded block gas limit in block headers is incorrect
The block gas limit is hardcoded to 21,000,000 in returned block headers. This value is incorrect.

https://github.com/oasisprotocol/emerald-web3-gateway/blob/50a5aefa81c2f9f8f3e5b2c4f8e0cec98d896ea2/indexer/utils.go#L88

ptrus:
Fixes: https://github.com/oasisprotocol/emerald-web3-gateway/issues/151

### DIFF
--- a/indexer/backend.go
+++ b/indexer/backend.go
@@ -77,6 +77,7 @@ type Backend interface {
 	Index(
 		oasisBlock *block.Block,
 		txResults []*client.TransactionWithResults,
+		blockGasLimit uint64,
 	) error
 
 	// Prune removes indexed data for rounds equal to or earlier than the passed round.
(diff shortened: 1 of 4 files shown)


indexer/backend.go
```diff
@@ -77,6 +77,7 @@ type Backend interface {
 	Index(
 		oasisBlock *block.Block,
 		txResults []*client.TransactionWithResults,
+		maxBatchSize uint64,
```
kostko: Why is this called `maxBatchSize`, probably better to call it `blockGasLimit`?

ptrus: renamed

kostko (approved):

PR #218, opened 2022-03-14
other(ci): remove not anymore needed hack

ptrus:
Merge once https://github.com/oasisprotocol/oasis-core/pull/4564 is released and the oasis-core dependency is bumped


ptrus:
will be done in https://github.com/oasisprotocol/emerald-web3-gateway/pull/225

ptrus:
nvm, this was not fixed properly, will need one more fix

PR #219, opened 2022-03-15
feat: add user-agent logging to key RPCs

nhynes:
This PR causes the user-agent to be logged at the info level for `sendRawTransaction`, `Call`, and `EstimateGas` RPCs. This will help us optimize the experience for the most common clients.

The `User-Agent` is added to the `Context` by [go-ethereum](https://github.com/ethereum/go-ethereum/blame/514ae7cfa35b9db671b8271419add728964f0ca6/rpc/http.go#L246-L247). The context key will change if the geth dep is ever updated.

### DIFF
--- a/rpc/eth/api.go
+++ b/rpc/eth/api.go
@@ -379,6 +379,7 @@ func (api *PublicAPI) Call(args utils.TransactionArgs, blockNrOrHash ethrpc.Bloc
 	}
 
 	logger.Debug("response", "args", args, "resp", res)
+	logger.Info("executed call", "user_agent", api.ctx.Value("User-Agent"))
 
 	return res, nil
 }
@@ -409,7 +410,10 @@ func (api *PublicAPI) SendRawTransaction(data hexutil.Bytes) (common.Hash, error
 		return ethTx.Hash(), err
 	}
 
-	return ethTx.Hash(), nil
+	txHash := ethTx.Hash()
+	logger.Info("submitted transaction", "user_agent", api.ctx.Value("User-Agent"), "tx_hash", txHash)
+
+	return txHash, nil
 }
 
 // EstimateGas returns an estimate of gas usage for the given transaction .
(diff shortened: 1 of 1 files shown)


rpc/eth/api.go
```diff
@@ -409,7 +410,10 @@ func (api *PublicAPI) SendRawTransaction(data hexutil.Bytes) (common.Hash, error
 		return ethTx.Hash(), err
 	}
 
-	return ethTx.Hash(), nil
+	txHash := ethTx.Hash()
+	logger.Info("submitted transaction", "user_agent", api.ctx.Value("User-Agent"), "tx_hash", txHash)
```
nhynes, line 414: To avoid polluting the logs, I intend to fetch the tx contents from the indexer later.

ptrus: Would it make more sense for these to be of Debug level?

nhynes: I wouldn't disagree, but then they have to be grepped out of a million other things including raw tx data, which might get costy (google stackdriver is $0.50/gb). What do you think? I'm fine making it Debug if you don't think this will make the useful metric useless in context.

ptrus: Hm ok, how about keeping these as Info but not merging this PR then? (can the gateways that will collect this be built manually with this patch?)? However, if that's too much of a hassle, I'm also fine with merging this.


Maybe a good long-term solution would be cleaning up the existing Debug logging and adding the user-agent logging as Debug logs. I think that things like raw-tx data is probably only ever useful for failed transactions no need to log it for every transaction.

ptrus (approved):

ptrus:
Approving, feel free to merge if adding these as Info saves a lot of time/costs vs other possible solutions.

nhynes:
算了吧。I'll just run it locally. I'm probably the only person who need this. Thanks for the review, though!

PR #238, opened 2022-03-30
indexer: add timeout to indexing a block

ptrus:
Add timeout to `indexBlock` as runtime queries can block forever if the oasis-node runtime is not working.

This makes it easier to debug the case, since the indexer will report runtime queries timing out.

### DIFF
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -22,14 +22,12 @@ const (
 	pruningCheckInterval             = 60 * time.Second
 	healthCheckInterval              = 10 * time.Second
 	healthCheckIndexerDriftThreshold = 10
+
+	blockIndexTimeout = 60 * time.Second
 )
 
-var (
-	ErrGetBlockFailed        = errors.New("get block failed")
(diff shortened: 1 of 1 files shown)


indexer/indexer.go
```diff
@@ -61,30 +59,30 @@ func (s *Service) Health() error {
...
 	}
 
-	txs, err := s.client.GetTransactionsWithResults(s.ctx, blk.Header.Round)
+	txs, err := s.client.GetTransactionsWithResults(ctx, blk.Header.Round)
 	if err != nil {
-		return ErrGetTransactionsFailed
+		return fmt.Errorf("querying transactions with results: %w", err)
 	}
 
 	err = s.backend.Index(blk, txs, s.blockGasLimit)
```
ptrus, line 84: Would be great if also this method also accepted a context, instead of using an internal one, but this fix will be done separate PR as it requires a bit of a bigger refactor.

kostko (approved):

PR #241, opened 2022-03-30
feat: implement monitoring for API endpoints

### LINKED ISSUE #230: Implement instrumentation
Add instrumentation (prometheus?) metrics for all endpoints. For each API endpoint, the number of invocations, number of failures and processing time should be measured. 
Could also do the same for all database queries.

ptrus:
Closes: https://github.com/oasisprotocol/emerald-web3-gateway/issues/230

TODO:
- [x] configurable
- [x] add some metrics to track request heights (difference from latest)
- [x] test

### DIFF
--- a/rpc/metrics/metrics.go
+++ b/rpc/metrics/metrics.go
@@ -1,0 +1,48 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	durations = promauto.NewHistogramVec(prometheus.HistogramOpts{Name: "oasis_emerald_web3_gateway_api_seconds", Buckets: []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10, 20, 30}, Help: "Histogram for the eth API requests duration."}, []string{"method_name"})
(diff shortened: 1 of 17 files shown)


rpc/metrics/metrics.go
```diff
@@ -0,0 +1,49 @@
...
+		inflight.WithLabelValues(method),
+		durations.WithLabelValues(method)
+}
+
+// InstrumentCaller instruments the caller method.
+//
+// Use InstrumentCaller is usually used the following way:
+//    func InstrumentMe() (err error) {
+// 	  r, s, f, i, d := metrics.GetAPIMethodMetrics("method")
+// 	  defer metrics.InstrumentCaller(r, s, f, i, d, err)()
```
kostko: Will this pass the error correctly?

ptrus: Good catch, thanks! 

As discussed added a pointer to error as an argument. Not the nicest/most conventional, but one added benefit is that now the `error` does not acctually need to be defined in the function argument scope for this to work.

One alternative would be splitting error monitoring functionality into a separate function and calling that from the defered function:
```golang
func InstrumentMe() (err error) {
  defer func(){
    metrics.InstrumentCallerErr(err)
  }()
  // Do actual work.
}
```
Can also switch to that if others prefer it.

kostko (approved):

PR #244, opened 2022-03-31
feat(indexer): Moar cache

Yawning:


ptrus:
Would adding one more cache for `GetLogs` (per round) make sense as well?

Yawning:
> Would adding one more cache for `GetLogs` (per round) make sense as well?

In theory, yes.  In practice, it depends on the access patterns, and how much work we want to put into it.  The API doesn't help here, because `GetLogs` takes a range.  If the state of the cache is pathologically sparse, then having a cache can end up using more database interactions than not having one.

My current inclination here is to wait till we have metrics on how often this is used, and what the requests look like.  If requests like `GetLogs(10, 10)` is common, then I would not be opposed to special casing that.

ptrus (approved):

PR #251, opened 2022-04-04
deps: Bump oasis-node and emerald-paratime in emerald-dev

### LINKED ISSUE #250: Bump emerald-dev image to oasis-node to 22.1+
Bump to oasis-core to at least 22.1and emerald-paratime 8.0.0-testnet. This requires adding `--debug.allow_root` flag to oasis-node to avoid the new EUID bit checks. It seems there's also some other issues with the bump.

matevz:
Fixes #250


ptrus:
Note that this won't work until a new version of the emerald paratime is released, which includes: https://github.com/oasisprotocol/oasis-sdk/pull/890

ptrus:
This was done in https://github.com/oasisprotocol/emerald-web3-gateway/pull/290

PR #256, opened 2022-04-07
feature: implement a gas price oracle based on indexed blocks

ptrus:


db/migrations/20220109122505_logs.up.go
```diff
@@ -44,7 +44,7 @@ type Log struct {
 	BlockHash string
 	TxHash    string `bun:",pk"`
 	TxIndex   uint
-	Index     uint `bun:",pk,allowzero"`
```
ptrus, line 47: This is a no-op change, as `allowzero` tag was removed in one of the past versions, so it is ignored anyway (and not needed)

gas/backend.go
```diff
@@ -0,0 +1,258 @@
...
+	service.BackgroundService
+
+	// GasPrice returns the currently recommended minimum gas price.
+	GasPrice() *hexutil.Big
+}
+
+const (
+	// windowSize is the number of recent blocks to use for calculating min gas price.
+	// NOTE: code assumes that this is relatively small.
+	windowSize = 3
```
kostko: This seems quite small, should we increase it to something like 12 blocks or so?

gas/backend.go
```diff
@@ -0,0 +1,258 @@
...
+
+var (
+	// minPriceEps is a constant added to the cheapest transaction executed in last windowSize blocks.
+	minPriceEps = *quantity.NewFromUint64(1_000_000_000) // 1 "gwei".
+
+	// defaultGasPrice is the default gas price reported if no better estimate can be returned.
+	//
+	// This is only returned when all calls to oasis-node for min gas price failed and there were no
+	// full blocks in last `windowSize` blocks.
+	defaultGasPrice = *quantity.NewFromUint64(10_000_000_000) // 10 "gwei".
```
kostko: Should this be updated to 100 gwei?

gas/backend.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	g.trackPrice(lastTxPrice)
+}
+
+func (g *gasPriceOracle) trackPrice(price *quantity.Quantity) {
+	// One item gets added to the prices array. Bump the current index
+	// for next iteration.
+	defer func() {
+		g.blockPricesCurrentIdx++
+		if g.blockPricesCurrentIdx >= windowSize {
```
kostko: Could use `g.blockPricesCurrentIdx = (g.blockPricesCurrentIdx + 1) % windowSize`.

indexer/utils.go
```diff
@@ -248,6 +258,11 @@ func (ib *indexBackend) StoreBlockData(ctx context.Context, oasisBlock *block.Bl
 			return err
 		}
 
+		// Remember the gas price as a fallback in case parsing the last transaction would fail for whetever reason.
```
kostko: This is not only a fallback, right? As if there are only Ethereum transactions, the above branch for the last transaction would not trigger.

tests/rpc/utils.go
```diff
@@ -56,14 +57,16 @@ type Response struct {
...
+	OasisBlockTimeout = 60 * time.Second
 	GasLimit          = uint64(1_000_000) // Minimum gas limit required to pass all tests (not using gas estimation).
 )
 
 var (
-	db       *psql.PostDB
-	GasPrice = big.NewInt(10_000_000_000) // Minimum gas price: https://github.com/oasisprotocol/emerald-paratime/blob/07179c1d20eddad3b3c3c9373bc0c853ad57fa16/src/lib.rs#L36
-	w3       *server.Web3Gateway
+	db             *psql.PostDB
+	GasPrice       = big.NewInt(10_000_000_000) // Minimum gas price: https://github.com/oasisprotocol/emerald-paratime/blob/07179c1d20eddad3b3c3c9373bc0c853ad57fa16/src/lib.rs#L36
```
kostko: Should this be updated?

gas/backend.go
ptrus: Agreed, maybe also the `fullblockThreshold` should be lowered a bit? Maybe 0.8?

kostko: Yeah sounds good.

tests/rpc/utils.go
ptrus: Updated, no harm in updating before bumping the runtime version (which will be done in a separate PR).

indexer/utils.go
ptrus: True. Reworded the comment a bit.

gas/backend.go
ptrus: Updated

kostko (approved):

Issue #252, opened 2022-04-05
feat(indexer): Improve cache efficiency


Yawning:
The caching added by #242 and #244 seems to be effective(?), but requires tuning and is likely massive overkill.  Based on metrics the hypothesis that the overwhelming majority of queries will be for temporally recent data is empirically true.

With the access patterns observed, it is likely that it is possible dramatically shrink and simplify the cache(s) like thus:

 - Cache data for the last `n` blocks, as they are indexed.
 - Hit up the database on cache miss, do not add the fetched data to the cache.

It is likely that `n` can be set rather small, but this somewhat depends on how the `getLogs` query is being used.  Perhaps #249 can add deltaStart/deltaEnd/rangeSize metrics for the call, if they are not present already.

PR #260, opened 2022-04-10
feat(indexer): Improve cache memory efficiency

Yawning:
Instead of using a complicated cache implementation, just use a handful
of sync.Maps along with a "remove eldest" replacement policy.  Since our
critical-path access patterns are overwhelmingly for temporally recent data
the new policy also makes it impossible for the hot dataset to be evicted.

Implements #252

### DIFF
--- a/indexer/backend.go
+++ b/indexer/backend.go
@@ -93,8 +93,9 @@ type Backend interface {
 
 // BlockData contains all per block indexed data.
 type BlockData struct {
-	Block    *model.Block
-	Receipts []*model.Receipt
+	Block      *model.Block
+	Receipts   []*model.Receipt
+	UniqueTxes []*model.Transaction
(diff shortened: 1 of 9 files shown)


Yawning:
While I am marking this ready for review, I'm not sure how much of a performance difference this will make.

ptrus (approved):

indexer/backend.go
```diff
@@ -93,8 +93,9 @@ type Backend interface {
 
 // BlockData contains all per block indexed data.
 type BlockData struct {
-	Block    *model.Block
-	Receipts []*model.Receipt
+	Block      *model.Block
+	Receipts   []*model.Receipt
+	UniqueTxes []*model.Transaction
```
ptrus, line 98: Oh, yeah this is correct, I shouldn't have changed this

ptrus (approved):

PR #268, opened 2022-04-26
feat: Add (partial) archive node support

Yawning:
* [x] Add a composite client
 * [x] Wire in the composite client

### DIFF
--- a/client/multi.go
+++ b/client/multi.go
@@ -1,0 +1,328 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"math"
+
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
+	"github.com/oasisprotocol/oasis-core/go/common/pubsub"
+	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
+	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/client"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/types"
+)
+
+func clientBlockRanges(
+	ctx context.Context,
+	cli client.RuntimeClient,
+) (uint64, uint64, error) {
+	blk, err := cli.GetLastRetainedBlock(ctx)
+	if err != nil {
+		return 0, 0, fmt.Errorf("client: failed to query last retained block: %w", err)
+	}
+	start := blk.Header.Round
+
+	blk, err = cli.GetBlock(ctx, client.RoundLatest)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


Yawning:
This maybe is enough to get historical EVM queries supported, but I have no idea if this actually works.  Someone that cares should test this, because I do not.

client/multi.go
```diff
@@ -0,0 +1,328 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("client: failed to query active rt info: %w", err)
+	}
+
+	mc := &MultiClient{
+		active:         active,
+		archives:       make([]*archiveClient, 0, len(archives)),
+		archivesEldest: math.MaxUint64,
+	}
+	if mc.archivesEldest, _, err = clientBlockRanges(ctx, active); err != nil {
```
ptrus: This should set `activeEldest` right? (not archive)

Yawning: yes

ptrus (approved):

Yawning:
Welp, this doesn't actually work at all because the client-sdk can only talk to the version of the oasis-node it was built against.  The ideal solution is to wait for someone (that's not me) to implement https://github.com/oasisprotocol/oasis-sdk/issues/954.

gw0:
An alternative solution is to import older versions of `client-sdk` by changing the Git versioning or by using the `replace` directive (https://www.percona.com/blog/2020/03/09/using-different-versions-of-a-package-in-an-application-via-go-modules/). Then for each archive node a different version should be used (either explicitly provided in the config or implicitly by querying the supported versions).

Yawning:
> An alternative solution is to import older versions of `client-sdk` by changing the Git versioning or by using the `replace` directive (https://www.percona.com/blog/2020/03/09/using-different-versions-of-a-package-in-an-application-via-go-modules/). Then for each archive node a different version should be used (either explicitly provided in the config or implicitly by querying the supported versions).

That is an objectively terrible solution because it relies on the public interfaces exported by the client sdk not changing at all between versions, or a mountain of code duplication in this codebase.  This doesn't actually work (easily anyway) either because `replace` isn't very good, and you get things like:

```
# github.com/oasisprotocol/oasis-core/go/v21/common/grpc
../../../go/pkg/mod/github.com/oasisprotocol/oasis-core/go@v0.2103.13/common/grpc/cbor.go:20:14: undefined: cbor.UnmarshalHigherLimits
```

Yawning:
Unless someone has better ideas, this is a total non-starter (even having the SDK magically handle backward compatibility is.... "hard if not way more trouble than it's worth"), so I am tempted to close this PR as a WONTFIX.

Yawning:
This approach has way too many problems, and making it work is 100% more trouble than it's worth.

Issue #84, opened 2021-11-30
Add an option to disable the indexer


ptrus:
That wold enable having multiple instances of web3 deployed using the same DB where one would be used as the indexer, and other would be used for serving queries (only need read-only access to the DB).

ptrus:
This would also require changing the event system, as events are sent directly from the indexer https://github.com/oasisprotocol/emerald-web3-gateway/blob/50a5aefa81c2f9f8f3e5b2c4f8e0cec98d896ea2/indexer/utils.go#L291-L297

PR #281, opened 2022-05-27
feat: add archive web3 gw support

Yawning:


Yawning:
This adds an indexer disable (See #84) but not the event related handling, so the option is only useful to allow archival instances to (in theory) share the db with a non-archival instance.

ptrus (approved):

PR #287, opened 2022-06-15
Deployment Docs for Gateway (and archive gateway)

ravenac95:
At this time this _does not_ include any instructions for sharing a postgres deployment. That still needs to undergo testing.


docs/deployment.md
```diff
@@ -0,0 +1,312 @@
+# Deploying the Emerald Web3 Gateway
+
+## Deployment Architecture
+
+To deploy an emerald web3 gateway, a single instance (e.g. VM, Bare Metal,
+Kubernetes Pod, etc.) of the deployment looks like this:
+
+![emerald web3 gateway instance](./gateway-instance.png)
```
ptrus, line 8: Hm, the `<latest_version>` part in the image might be confusing. Staring with the latest few emerald releases, there is a table now at which release that shows compatible oasis-node and emerald-paratime versions.

ptrus: I would also write "emerald paratime" to the oasis-node box in the image. (since the node needs also to run the emerald paratime)

docs/deployment.md
```diff
@@ -0,0 +1,312 @@
...
+time, does not go over using that configuration._
+
+As is shown in the above diagram, an emerald web3 gateway currently requires a
+colocated oasis-node and (at this time) a colocated postgres server. Not shown
+in the diagram is any load balancing infrastructure, that part of the
+architecture is left up to the reader.
+
+### Archive architecture
+
+In order to support archive emerald web3 requests, the following architecture is
```
ptrus, line 21: Maybe add a sentence explaining what archive is/when it's needed. (e.g. it's needed if you'd like to support pre-network upgrade state)

docs/deployment.md
```diff
@@ -0,0 +1,312 @@
...
+
+For archive support, multiple versions of an emerald web3 gateway deployment are
+required and are linked together in a chain. The top of the chain is intended to
+be the most current web3 gateway node representing the present state of the
+network. The bottom of the chain would be the oldest archive node. At this time
+there are only the cobalt and damask nodes that can be run in this way, so 2
+layers of web3 gateway nodes are required. In the future, multiple layers of
+web3 gateways may exist.
+
+## Deploying the Damask (oasis-node v22) Emerald Web3 gateway
```
ptrus, line 34: Maybe the "Deploying the Damask" section should be before the "archive mode" section.

ptrus (approved):
Looks good!

Left some minor comments, no need to address all of them.

docs/deployment.md
```diff
@@ -0,0 +1,312 @@
...
+Kubernetes Pod, etc.) of the deployment looks like this:
+
+![emerald web3 gateway instance](./gateway-instance.png)
+
+_Note: the latest version of the web3 gateway server has an untested
+configuration that allows for a shared postgres database. This guide, at this
+time, does not go over using that configuration._
+
+As is shown in the above diagram, an emerald web3 gateway currently requires a
+colocated oasis-node and (at this time) a colocated postgres server. Not shown
```
ptrus, line 15: Does the gateway really require a colocated postgres server? I think it does not

ravenac95: Ah ya you're right it wouldn't be _required_.

PR #308, opened 2022-08-04
feat: add oasis_publicKey RPC

nhynes:
This PR adds an optional method `oasis_publicKey` that returns the attached runtime's calldata public key. This is helpful for Web3 clients that don't want to also connect directly to GRPC.

### DIFF
--- a/rpc/oasis/api.go
+++ b/rpc/oasis/api.go
@@ -1,0 +1,41 @@
+package oasis
+
+import (
+	"context"
+	"errors"
+
+	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/client"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/core"
+)
+
+var ErrInternalError = errors.New("internal error")
+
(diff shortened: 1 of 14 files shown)


rpc/oasis/api.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/client"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/core"
+)
+
+var ErrInternalError = errors.New("internal error")
+
+// API is the net_ prefixed set of APIs in the Web3 JSON-RPC spec.
+type API interface {
+	// PublicKey returns the calldata public key for the runtime with the provided ID.
+	PublicKey(ctx context.Context, runtimeID hexutil.Bytes) (*core.CallDataPublicKeyResponse, error)
```
kostko: Why does it need the `runtimeID`? The whole gateway supports a specific runtime so this shouldn't be needed.

nhynes: If I hadn't logged the variable, the lint would have caused me to realize this :)

kostko:
Maybe we should call it `oasis_callDataPublicKey` just so that it is explicit what the public key is for in case we once decide to have other keys as well.

kostko (approved):

PR #340, opened 2022-12-09
Drop unencrypted transactions by default

matevz:
This PR:
- drops unencrypted eth transaction, if Oasis RPC is enabled (e.g. sapphire) and the data is a wrapped Oasis transaction with CallFormat==Plain
- adds `gateway.allow_unencrypted_txs` flag to config which enables accepting unencrypted transactions
- adds `conf/tests-c10l.yml` and use it in Sapphire e2e tests
- enables Oasis RPC in default `server.yml` config

### DIFF
--- a/rpc/eth/api.go
+++ b/rpc/eth/api.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ethereum/go-ethereum/eth/filters"
 	"github.com/ethereum/go-ethereum/rlp"
 	ethrpc "github.com/ethereum/go-ethereum/rpc"
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/client"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/crypto/signature/secp256k1"
@@ -108,13 +109,14 @@ type API interface {
 }
 
 type publicAPI struct {
-	client         client.RuntimeClient
-	archiveClient  *archive.Client
-	backend        indexer.Backend
-	gasPriceOracle gas.Backend
-	chainID        uint32
-	Logger         *logging.Logger
-	methodLimits   *conf.MethodLimits
+	client              client.RuntimeClient
+	archiveClient       *archive.Client
+	backend             indexer.Backend
+	gasPriceOracle      gas.Backend
+	chainID             uint32
+	Logger              *logging.Logger
+	methodLimits        *conf.MethodLimits
+	allowUnencryptedTxs bool
 }
 
 // NewPublicAPI creates an instance of the public ETH Web3 API.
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-test.yaml
```diff
@@ -142,7 +142,9 @@ jobs:
         run: tests/tools/spinup-oasis-stack.sh > /dev/null &
 
       - name: Unit tests with coverage
-        run: go test -race -coverpkg=./... -coverprofile=coverage.txt -covermode=atomic -v ./...
+        run: |
+          cp conf/tests-c10l.yml conf/tests.yml
```
ptrus: Maybe it would be nicer to use the default test config (`conf/tests.yml`) and use ENV variables to set the necessary overrides.

rpc/eth/api.go
```diff
@@ -463,6 +467,13 @@ func (api *publicAPI) SendRawTransaction(ctx context.Context, data hexutil.Bytes
 		return common.Hash{}, ErrMalformedTransaction
 	}
 
+	if !api.checkOasisTxEncrypted(ethTx.Data()) {
+		logger.Debug("dropped unencrypted transaction", "hash", ethTx.Hash())
+		return common.Hash{}, ErrInvalidRequest
+	}
+
+	ethTx.Data()
```
ptrus, line 475: This looks like a leftover?

rpc/eth/api.go
```diff
@@ -463,6 +467,13 @@ func (api *publicAPI) SendRawTransaction(ctx context.Context, data hexutil.Bytes
 		return common.Hash{}, ErrMalformedTransaction
 	}
 
+	if !api.checkOasisTxEncrypted(ethTx.Data()) {
```
ptrus, line 470: Maybe rename the method to `isEncrypted(tx)`  and only call it if `allowUnencryptedTxs` is set to false.  Then it can also be made a standalone function (and not a method on the api backend), which will make it easy to write some unit tests for it. 
It might also make sense to return a more descriptive error on rejection. 

```golang
if !api.allowUnencryptedTxs && !isEncrypted(ethTx.Data()) {
    logger.Debug("unencrypted transaction rejected", "hash" ..., 
    return common.Hash{}, ErrUnencryptedNotAllowed
}   
```

kostko:
Hm, I'm not convinced this belongs in the gateway? Shouldn't this be up to the clients?

rpc/eth/api.go
```diff
@@ -751,3 +762,28 @@ func (api *publicAPI) getBlockRound(ctx context.Context, logger *logging.Logger,
...
+	if api.allowUnencryptedTxs {
+		// Unencrypted transactions are allowed or encryption not supported by the gateway.
+		return true
+	}
+	if data == nil {
+		// Transaction is not Oasis transaction, ignore.
+		return true
+	}
+
+	var tx types.Transaction
```
kostko, line 777: This does not seem correct as the `data` field does not contain an entire SDK transaction, but rather just a call? This should have a corresponding E2E test.

CedarMist:
Re: discussion with Matevz

Based on the hostname, could we force encrypted transaction?

E.g. `sapphire-encrypted-only-rpc.oasis.io`

matevz:
> Re: discussion with Matevz
> 
> Based on the hostname, could we force encrypted transaction?
> 
> E.g. `sapphire-encrypted-only-rpc.oasis.io`

To put our discussion into context, since we still want to support standard Ethereum transactions, we cannot simply drop all unencrypted ones on the existing public RPC endpoint. But what we can do for c10l dApp developers out there that know they want to use explicitly encrypted txes and queries is to set up **another** Web3 endpoint that **requires** encrypted transactions, e.g. encrypted.sapphire.oasis.io.

While this is fine for CLI dApps, where you simply point to a different RPC, it may cause some issues on the frontend MetaMask where the RPC URL should be unique per chain ID.

PR #341, opened 2022-12-11
breaks: rename to oasis-web3-gateway

ptrus:
TODO:
- [x] once this is merged, rename the repository on github to: oasis-web3-gateway

The code was not tied to the emerald paratime, so there should be no breaking changes in the configs.

Still marking the change as breaking, as anyone using the code as a library would have breaking changes.


kostko (approved):
He he, evm-web3-gateway -> emerald-web3-gateway -> oasis-web3-gateway :tada:

Issue #411, opened 2023-08-09
Can we reduce the startup time of the sapphire-dev container?


CedarMist:
Currently with the `mock` backend it takes about a minute for sapphire-dev container to start on my laptop.

Ideally it should take less time, so it's worth doing some investigation to determine why it takes so long.

We probably shouldn't try things blindly, and instead try to instrument the container startup.

Some options are (unsure if they would make any difference):

  - [x] Reduce logging? (maybe disk I/O is causing delays?)
  - [x] Run only 1 compute node if possible.
  - [x] Removing additional sleeps
  - [x] Provision less accounts

There are ways of tracking startup time (e.g. a flame graph) which could be useful:

 - https://github.com/mboussaa/docker-flame-graphs
 - https://www.intel.com/content/www/us/en/docs/vtune-profiler/cookbook/2023-1/profiling-in-docker-container.html
 - https://stackoverflow.com/questions/53698972/how-to-do-profiling-inside-a-docker-container
 - https://nanxiao.me/en/use-perf-and-flamegraph-to-profile-program-on-linux/
 - https://gendignoux.com/blog/2019/11/09/profiling-rust-docker-perf.html

ptrus:
Could you post some logs (with timestamps) of the startup part. I assume it would be somewhat easy to see what steps are using most of the time and how realistically it is reducing it down.

> Reduce logging? (maybe disk I/O is causing delays?)

Most likely not the case.

> Run only 1 compute node if possible.

Yeah should be possible and easy to setup via the fixture (I think the default uses 3 compute nodes).

> Provision less accounts

How many accounts are currently provisioned? I assume all accounts are provisioned from the same "seed account" so it can only provision one new account per runtime block, so this could indeed slow the startup down. This could be optimized. 

---

Some more ideas:
- if we find out there's no way to meaningfully reduce the startup-time further, the start-up phase could be moved into the build step of the docker container. We could, as part of the docker build phase, start and initialize the network state. On container startup it would then just need to resume the already initialized network - which should be fast (note that the oasis-node net-runner does already support this - resuming an already provisioned network). 

- related to the above. Make the container non-ephemeral. With some updates the container could persist state across restarts, removing the need to re-provision/initialize the network on every restart.

CedarMist:
I tend to run the containers with `--rm` otherwise I end up with a bazillion previously used Docker containers which makes it difficult to prune unused stuff in the docker cache.

[...]

PR #413, opened 2023-08-14
sapphire-dev: deposit test mnemonic, core-22.2.11, optionally use single compute note

CedarMist:
This makes 4 changes:

 - Increase number of tokens to 100,000 in `oasis-deposit`
 - Add `-test-mnemonic` flag to `oasis-deposit` which uses the `test test test ... junk` mnemonic
    - This addresses a problem I was having with CI in sapphire-paratime, couldn't get the `-to` flag working with a mnemonic via GitHub workflow (no combination of quoting or escaping would let me pass a quoted parameter with spaces in it!)
 - Upgrade to Oasis Core 22.2.11 
 - Use a single compute node by default when running `sapphire-dev` (reduces CPU usage)

Additionally there was some weirdness with PostgreSQL I had when building the `sapphire-dev` container that I didn't have early last week and was unable to find the root cause. After installing the postgresql package some of the permissions were wrong which prevented it from starting. Running `chown` wouldn't persist across `RUN` statements, meaning I had to prefix the `/etc/init.d/postgresql start` command with `chown`, and then run it again in `start.sh` upon startup. I'm using Docker `20.10.25-0ubuntu1~22.04.1`.

Re #411: using 1 compute node instead of 2 doesn't significantly affect startup time, but it does help battery life. Am still (slowly) investigating idle CPU usage and startup time (wait-node)

The startup process now looks like:

```
docker run --rm -it -p8545:8545 -p8546:8546 ghcr.io/oasisprotocol/sapphire-dev:local -test-mnemonic
sapphire-dev local (oasis-core: 22.2.11, sapphire-paratime: 0.5.2, oasis-web3-gateway: 3.3.0-git0d09728+dirty)

 * Starting oasis-net-runner with sapphire
 * Starting PostgreSQL 14 database server                                                                             [ OK ] 
 * Starting oasis-web3-gateway
 * Bootstrapping network (this might take a minute)...
 * Populating accounts

Available Accounts
==================
(0) 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 (100000 TEST)
(1) 0x70997970C51812dc3A010C7d01b50e0d17dc79C8 (100000 TEST)
(2) 0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC (100000 TEST)
(3) 0x90F79bf6EB2c4f870365E785982E1f101E93b906 (100000 TEST)
(4) 0x15d34AAf54267DB7D7c367839AAf71A00a2C6A65 (100000 TEST)

Private Keys
==================
(0) 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
(1) 0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d
(2) 0x5de4111afa1a4b94908f83103eb1f1706367c2e68ca870fc3fb9a804cdab365a
(3) 0x7c852118294e51e653712a81e05800f419141751be58f605c371e15141b007a6
(4) 0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a

HD Wallet
==================
Mnemonic:	test test test test test test test test test test test junk
Base HD Path:	m/44'/60'/0'/0/%d

WARNING: The chain is running in ephemeral mode. State will be lost after restart!

Listening on http://localhost:8545 and ws://localhost:8546
```


nhynes (approved):

tests/tools/spinup-oasis-stack.sh
```diff
@@ -11,6 +11,7 @@ set -euo pipefail
 # - PARATIME_VERSION: version of the binary. e.g. 3.0.0
 # - OASIS_NODE_DATADIR: path to temporary oasis-node data dir e.g. /tmp/oasis-localnet
 # - KEYMANAGER_BINARY: path to key manager binary e.g. simple-keymanager
+# - OASIS_SINGLE_COMPUTE_NODE: Only run a single compute note
```
kostko:
```suggestion
# - OASIS_SINGLE_COMPUTE_NODE: Only run a single compute node
```

kostko (approved):

PR #424, opened 2023-09-04
feat: Add support for Sapphire 0.6.2-testnet

kostko:
Uses Sapphire 0.6.2-testnet for the sapphire-dev container


CedarMist (approved):
This passes all my unit tests.

There is still the outstanding issue with 4second delay in web3 gateway, but I will look into this tomorrow.

PR #431, opened 2023-09-11
fix: oasis deposit multiple accounts

### LINKED ISSUE #430: -test-mnemonic -n does not work
When using `-test-mnemonic -n 5` in sapphire-dev Docker image it only funds the very first account (0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266) although the next four should also be funded (writes on stdout though). Fix the `-n` parameter to work correctly.

matevz:
Fixes https://github.com/oasisprotocol/oasis-web3-gateway/issues/430

This PR:
- bumps oasis-sdk go client to 0.6.1
- increases the balance of `test:alice`. It's now funded with 1M tokens. Running out of funds was also the reason for bug #430.
- adds support for allowance transactions in oasis-deposit (allowance is currently defined in the genesis for faster startup, so this is not used for now)
- reduces funding of accounts to 10,000 tokens by default so we can fund 20 accounts. This is also in line with `npx hardhat node`
- adds support for multiple `-to` comma-separated addresses


docker/common/oasis-deposit/main.go
```diff
@@ -241,9 +300,8 @@ func main() {
 		} else if err = addr.UnmarshalText([]byte(a)); err != nil {
 			panic(fmt.Sprintf("unmarshal addr err: %s", err))
 		}
-
-		txb := consAcc.Deposit(&addr, baseAmount).SetFeeConsensusMessages(1)
-		_, err = SignAndSubmitTx(ctx, rtc, testing.Alice.Signer, *txb.GetTransaction(), 0)
+		txb := consAcc.Deposit(&addr, baseAmount).SetFeeConsensusMessages(5)
```
kostko: Why 5 consensus messages? I don't think this is needed as only a single deposit happens in this transaction.

docker/common/oasis-deposit/main.go
```diff
@@ -81,6 +97,27 @@ func EstimateGas(ctx context.Context, rtc client.RuntimeClient, tx types.Transac
 	return tx
 }
 
+// SignAndSubmitConsensusTx signs and submits the given transaction.
+// Gas estimation is done automatically.
+func SignAndSubmitConsensusTx(ctx context.Context,
+	cs consensus.ClientBackend,
+	signer coreSignature.Signer,
+	tx *transaction.Transaction) error {
```
kostko:
```suggestion
	tx *transaction.Transaction,
) error {
```

docker/common/oasis-deposit/main.go
```diff
@@ -81,6 +97,27 @@ func EstimateGas(ctx context.Context, rtc client.RuntimeClient, tx types.Transac
 	return tx
 }
 
+// SignAndSubmitConsensusTx signs and submits the given transaction.
+// Gas estimation is done automatically.
+func SignAndSubmitConsensusTx(ctx context.Context,
```
kostko:
```suggestion
func SignAndSubmitConsensusTx(
	ctx context.Context,
```

kostko (approved):

docker/common/oasis-deposit/main.go
matevz: Sorry, debugging leftover.

PR #447, opened 2023-09-20
feat: Support simplified revert reason encoding

### LINKED ISSUE #429: Custom Solidity errors are not passed correctly
Using custom `error` type with `revert` in Solidity isn't passed correctly to the client.

```solidity
    error Blah(uint);
    function testRevert () external view {
        revert Blah(123);
    }
```

The Ethers v5 TypeChain provider will automatically decode errors, the code should return this as the `data`:

    0x966682ee000000000000000000000000000000000000000000000000000000000000007b

However, with Sapphire 0.6.2-testnet, it returns this instead:

    0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000306c6d6143376741414141414141414141414141414141414141414141414141414141414141414141414141414141423700000000000000000000000000000000

This is due to the error handler which re-encodes the errors:

https://github.com/oasisprotocol/oasis-web3-gateway/blob/285aefcffb7e65977f0d57fc9906c8d7fce6b95a/rpc/eth/api.go#L368-L382

The web3-gateway log for this is:

```json
{"caller":"grpc.go:242","err":"reverted: lmaC7gAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB7","level":"error","method":"/oasis-core.RuntimeClient/Query","module":"grpc/client","msg":"request failed","req_seq":310310,"rsp":{"data":null},"ts":"2023-09-08T02:37:08.654508996Z"}
```

Which is handled by: 

https://github.com/oasisprotocol/oasis-web3-gateway/blob/285aefcffb7e65977f0d57fc9906c8d7fce6b95a/rpc/eth/api.go#L440-L446

Ethers v5 predictably decodes this correctly as: 

```
Error: call revert exception; VM Exception while processing transaction: reverted with reason string "lmaC7gAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB7" [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="testRevert()", data="0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000306c6d6143376741414141414141414141414141414141414141414141414141414141414141414141414141414141423700000000000000000000000000000000", errorArgs=["lmaC7gAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB7"], errorName="Error", errorSignature="Error(string)", reason="lmaC7gAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB7", code=CALL_EXCEPTION, version=abi/5.7.0)
```

ptrus:
Closes: https://github.com/oasisprotocol/oasis-web3-gateway/issues/429

Relevant: https://github.com/oasisprotocol/oasis-sdk/pull/1479

Tested with the example in https://github.com/oasisprotocol/oasis-web3-gateway/issues/429 and manually built Sapphire-0.6.3:
```
error: {
  message: 'reverted: lmaC7gAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB7'
}
```

### DIFF
--- a/rpc/eth/revert_errors.go
+++ b/rpc/eth/revert_errors.go
@@ -1,0 +1,104 @@
+package eth
+
+import (
+	"context"
+	"encoding/base64"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/core"
+	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/evm"
+)
+
+const (
+	revertErrorPrefix = "reverted: "
+)
+
+type RevertError struct {
+	error
+	Reason string `json:"reason"`
+}
+
+// ErrorData returns the ABI encoded error reason.
+func (e *RevertError) ErrorData() interface{} {
+	return e.Reason
+}
+
+func (api *publicAPI) newRevertErrorV1(revertErr error) *RevertError {
+	// ABI encoded function.
+	abiReason := []byte{0x08, 0xc3, 0x79, 0xa0} // Keccak256("Error(string)")
+
+	// ABI encode the revert Reason string.
+	revertReason := strings.TrimPrefix(revertErr.Error(), revertErrorPrefix)
+	typ, _ := abi.NewType("string", "", nil)
+	unpacked, err := (abi.Arguments{{Type: typ}}).Pack(revertReason)
+	if err != nil {
+		api.Logger.Error("failed to encode V1 revert error", "revert_reason", revertReason, "err", err)
+		return &RevertError{
+			error: revertErr,
+		}
+	}
+	abiReason = append(abiReason, unpacked...)
+
+	return &RevertError{
+		error:  revertErr,
+		Reason: hexutil.Encode(abiReason),
+	}
+}
+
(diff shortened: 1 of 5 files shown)


rpc/eth/revert_errors.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// In EVM module version 1 the SDK decoded and returned readable revert reasons.
+	// Since EVM version 2 the raw data is just base64 encoded and returned.
+	// Changed in https://github.com/oasisprotocol/oasis-sdk/pull/1479
+	switch {
+	case rtInfo == nil || rtInfo.Modules[evm.ModuleName].Version > 1:
+		// EVM version 2 onward (if no runtime info assume >1).
+		logger.Debug("failed to execute SimulateCall", "err", err)
+		return err
```
kostko: Hm, I think this should still decode the error by stripping the revert prefix and base64 decoding the reason, then re-encoding it in hex and returning it as this special error with a `Reason` field. Otherwise this will not be handled properly by developer tools.

ptrus: You're right, got confused by a part of the #429 and thought that `reverted: lmaC7gAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAB7` is the expected response.

That makes more sense, updated the code. Will re-test.

kostko (approved):

PR #463, opened 2023-10-17
feat(docker): sapphire: reduce docker layers + 0.6.4 mainnet

CedarMist:
Upgrades the `sapphire-dev` container to 0.6.4 versus 0.6.4-testnet

The `Dockerfile` has been reorganized:
 * Reduce the number of layers by combining commands
 * Remove downloaded tar files and unused extracted files

Additionally `start.sh`, when used with the `mock` backend, will increment the epoch every 10 minutes - now queries current epoch before advancing, instead of fetching once and incrementing regardless of other things advancing the epoch.

This reduces the docker image size by approx 125mb

### DIFF
--- a/docker/common/start.sh
+++ b/docker/common/start.sh
@@ -80,11 +80,11 @@ echo "Listening on http://localhost:8545 and ws://localhost:8546"
 
 if [[ ${SAPPHIRE_BACKEND} == 'mock' ]]; then
 	# Run background task to switch epochs every 5 minutes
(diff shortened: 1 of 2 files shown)


ptrus (approved):
:+1: 
Too bad there's still no CI that would test the image

CedarMist:
@ptrus We can talk next week about the bigger picture of testing & circular dependencies etc. and how to untangle.

PR #469, opened 2023-10-25
fix: sapphire-dev: Fix key manager initialization

matevz:
This PR:
- waits for the key manager to generate the ephemeral secret (the first possible epoch for this is 3)  and then transitions to that epoch so the ephemeral key can be used by dApps (this behavior was new in Oasis Core 23.x)
- sets the beacon interval to 2. This interval determines among how many KM nodes the KM will propose the new ephemeral key. The previous value was 50 meaning there is 1/50 chance KM proposing the ephemeral secret every block, the block time is 1 second. On the Mainnet/Testnet this number is computed by taking the epoch length into account, but since we use mock epochs in the Docker image, the beacon interval needs to be hardcoded.
- bumps ci-test to use the new sapphire-runtime 7.0.0-testnet and emerald-runtime 11.0.0-testnet
- uses the **localnet**  version of emerald-runtime in ci-test as well for consistency with sapphire
- adds a test to sapphire-dev and emerald-dev for `eth_getBalance` which checks if the testing account was funded
- **emerald-dev tests currently disabled due to #471**
- adds a test for sapphire-dev, if `oasis_callDataPublicKey` works
- builds Docker images and tests them on pull requests too now, not just when pushed to main

### DIFF
--- a/docker/common/start.sh
+++ b/docker/common/start.sh
@@ -25,6 +25,7 @@ export SAPPHIRE_BACKEND=default
 fi
 
 OASIS_NODE_SOCKET=${OASIS_NODE_DATADIR}/net-runner/network/client-0/internal.sock
+OASIS_KM_SOCKET=${OASIS_NODE_DATADIR}/net-runner/network/keymanager-0/internal.sock
 
 set -euo pipefail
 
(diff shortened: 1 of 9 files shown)


docker/common/start.sh
```diff
@@ -63,6 +64,13 @@ if [[ ${SAPPHIRE_BACKEND} == 'mock' ]]; then
 
 	echo -n .
 	${OASIS_NODE} debug control set-epoch --epoch 2 -a unix:${OASIS_NODE_SOCKET}
+
+  # Transition to the final epoch when the KM generates ephemeral secret.
```
kostko: Fix indentation.

kostko (approved):

CedarMist (approved):

ptrus (approved):

peternose:
In E2E tests, the first ephemeral secret is available in epoch 2, not in epoch 3.
- Epoch 0: 
  - Key manager nodes register.
- Epoch 1: 
  - Key manager status changes, nodes are added to the committee. 
   - All nodes try to generate ephemeral secret for the next epoch. The first secret published is accepted, others rejected.
   - All nodes send the secret to the key manager enclave which verifies and stores the secret in memory.
- Epoch 2: 
  - Ephemeral secret for epoch 2 is available, keys can be derived. 
  - All nodes try to generate ephemeral secret for the next epoch...

PR #482, opened 2023-11-28
Reduce size of dev docker images

abukosek:
The size of the development docker images was reduced.
`sapphire-dev` was reduced from 835MB to 430MB.
`emerald-dev` was reduced from 683MB to 419MB.

### DIFF
--- a/.github/workflows/docker-dev.yml
+++ b/.github/workflows/docker-dev.yml
@@ -48,7 +48,7 @@ jobs:
           load: true
 
       - name: Test
-        timeout-minutes: 2
+        timeout-minutes: 3
         working-directory: docker/${{ matrix.docker_image }}
         run: ./test.sh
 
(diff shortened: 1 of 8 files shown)


CedarMist (approved):
Awesome :D 

```
sapphire-dev local (oasis-core: 23.0.6, sapphire-paratime: 0.7.0-testnet, oasis-web3-gateway: 4.0.0-git3c01224)
...
 * Container start-up took 51 seconds.
```

abukosek:
Thanks :D

CedarMist:
While this suggestion isn't blocking, in the sapphire-paratime repo we're using an additional script to wait until the key manager is fully functional, otherwise tests will fail if the container is used too early:

https://github.com/oasisprotocol/sapphire-paratime/blob/main/.github/wait-until-ready.sh

If a 'wait until ready' script was added to the root of the docker container, it would make CI easier as we could `docker exec` it to be sure the container is ready before proceeding.

abukosek:
That's a good idea -- I can make another PR that adds this.

PR #484, opened 2023-11-29
fix: Fix running start.sh without a terminal

abukosek:


CedarMist (approved):
I have tested locally without the Docker `-t` option and it runs fine without a tty.

This will fix CI tests on sapphire-paratime

PR #485, opened 2023-11-30
feat: build oasis-node and oasis-net-runner from source

CedarMist:
@abukosek if you're interested, this change builds oasis-core from source rather than pulling from github releases

I think we should have the option to build Sapphire paratime binary from source too as it will make testing changes quicker.

However, I think there should probably be two modes:
 * release mode - uses files downloaded from github releases
 * dev mode - builds everything from source

Will investigate if this is possible.


abukosek:
Cool!
Yeah, I also think there should be two separate modes for this.  The simplest way to do this would be to add another `ENV` var and then conditionally build from source or download existing artifacts based on that (or even better, an `ARG`, which we could then specify when building the image with `--build-arg`).

matevz:
Superseeded by https://github.com/oasisprotocol/oasis-web3-gateway/pull/601

PR #488, opened 2023-12-06
fix: Fix Oasis CLI in sapphire-dev Docker image

### LINKED ISSUE #487: alpine based docker can't run the bundled Oasis CLI
```
# ./oasis --help
Error relocating /oasis: __wcsncpy_chk: symbol not found

# ldd oasis
	/lib64/ld-linux-x86-64.so.2 (0x7f87c2905000)
	libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f87c2905000)
Error relocating oasis: __snprintf_chk: symbol not found
Error relocating oasis: __vfprintf_chk: symbol not found
Error relocating oasis: __open_2: symbol not found
Error relocating oasis: __pthread_unregister_cancel: symbol not found
Error relocating oasis: __poll_chk: symbol not found
Error relocating oasis: __asprintf_chk: symbol not found
Error relocating oasis: __wcsncpy_chk: symbol not found
Error relocating oasis: __vsnprintf_chk: symbol not found
Error relocating oasis: __fprintf_chk: symbol not found
Error relocating oasis: __pthread_register_cancel: symbol not found
```

abukosek:
Fixes #487.

~Note that in the long-term this should really be fixed in the CLI repo instead :)~ Apparently, this will break Ledger support, so we must keep this workaround just inside the Docker images, as they don't need Ledger support.

### DIFF
--- a/docker/sapphire-dev/Dockerfile
+++ b/docker/sapphire-dev/Dockerfile
@@ -8,10 +8,15 @@ RUN cd oasis-web3-gateway && make && strip -S -x oasis-web3-gateway docker/commo
 FROM ghcr.io/oasisprotocol/oasis-core-dev:stable-23.0.x AS oasis-core-dev
 
 ENV OASIS_UNSAFE_SKIP_KM_POLICY=1
+ENV OASIS_CLI_VERSION=0.7.1
 
 RUN git clone https://github.com/oasisprotocol/oasis-core.git --branch stable/23.0.x --depth 1 \
 	&& cd oasis-core/tests/runtimes/simple-keymanager \
-	&& cargo build --release
+	&& cargo build --release \
+	&& cd ../../../../ \
+	&& git clone https://github.com/oasisprotocol/cli.git --branch v${OASIS_CLI_VERSION} --depth 1 \
+	&& cd cli \
+	&& make CGO_ENABLED=0
 
 # Build sapphire-dev
 FROM postgres:16-alpine
@@ -20,7 +25,6 @@ RUN apk add --no-cache bash gcompat libseccomp jq binutils \
 
 # Docker-specific variables
 ENV OASIS_CORE_VERSION=23.0.9
-ENV OASIS_CLI_VERSION=0.6.0
 ENV PARATIME_VERSION=0.7.0-testnet
 ENV PARATIME_NAME=sapphire
 ENV GATEWAY__CHAIN_ID=0x5afd
@@ -42,6 +46,9 @@ ARG VERSION
 
 # simple-keymanager
 COPY --from=oasis-core-dev /oasis-core/target/release/simple-keymanager ${KEYMANAGER_BINARY}
+
+# cli
+COPY --from=oasis-core-dev /cli/oasis /oasis
 
 # oasis-web3-gateway binary, config, spinup-* scripts and staking_genesis.json.
 COPY --from=oasis-web3-gateway /go/oasis-web3-gateway/oasis-web3-gateway ${OASIS_WEB3_GATEWAY}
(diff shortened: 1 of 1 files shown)


kostko:
> Note that in the long-term this should really be fixed in the CLI repo instead :)

Building with `CGO_ENABLED=0` will break Ledger support in the CLI AFAIK in which case this is not doable.

docker/sapphire-dev/Dockerfile
```diff
@@ -20,7 +24,7 @@ RUN apk add --no-cache bash gcompat libseccomp jq binutils \
 
 # Docker-specific variables
 ENV OASIS_CORE_VERSION=23.0.9
-ENV OASIS_CLI_VERSION=0.6.0
+ENV OASIS_CLI_VERSION=master
```
kostko: Why master? There is 0.7.1 available.

abukosek: Because it builds the master branch?

abukosek:
> Building with `CGO_ENABLED=0` will break Ledger support in the CLI AFAIK in which case this is not doable.

Oh, OK, we can keep this workaround just here then.

docker/sapphire-dev/Dockerfile
abukosek: Oh, I see, the releases have tags, I'll change it to that then :)

abukosek: Fixed :)

abukosek:
Do we need Oasis CLI to work in the `emerald-dev` Docker image as well?

CedarMist:
I don't think many (any?) people are using the Emerald container, and if none of the CI tests are using it then I'd say no.

Likewise with Ledger support, it's highly unlikely that Ledger will be used with the Oasis CLI within the sapphire-dev docker container.

docker/sapphire-dev/Dockerfile
```diff
@@ -11,7 +11,11 @@ ENV OASIS_UNSAFE_SKIP_KM_POLICY=1
 
 RUN git clone https://github.com/oasisprotocol/oasis-core.git --branch stable/23.0.x --depth 1 \
 	&& cd oasis-core/tests/runtimes/simple-keymanager \
-	&& cargo build --release
+	&& cargo build --release \
+	&& cd ../../../../ \
+	&& git clone https://github.com/oasisprotocol/cli.git --branch v0.7.1 --depth 1 \
```
CedarMist: Is it possible to parameterize this? So we only need to update `OASIS_CLI_VERSION`

abukosek: I tried moving the `ENV` vars to the top, but Docker didn't like that, so I guess we'd have to use a separate `ARG` and pass the version when building the image.  I'd save this change for a later PR, though, because we can do it at the same time as adding support for building everything from git (we could then specify the versions in the Makefile).

kostko: Did you move everything or just the CLI version? I think one issue is that there are multiple builds defined in this Dockerfile and an ENV only applies to a single build (starting with FROM). But if the CLI version is no longer used below, moving it up (where the "OASIS_UNSAFE_SKIP_KM_POLICY" is) should work?

abukosek: You're right, that does work -- I'll push the changes now.

CedarMist (approved):
This works fine :)

```
# /oasis --version
Software version: v0.7.1-gitbcbf421
Oasis SDK version: v0.7.1
Oasis Core version: v0.2300.9
Go toolchain version: go1.21.3
```

```
sapphire-dev local (oasis-core: 23.0.9, sapphire-paratime: 0.7.0-testnet, oasis-web3-gateway: 4.0.0-git6b899c7)
```

I guess we can do the version bump to Sapphire 0.7.0 (not testnet) too?
 - https://github.com/oasisprotocol/sapphire-paratime/releases/tag/v0.7.0

docker/sapphire-dev/Dockerfile
```diff
@@ -20,7 +25,6 @@ RUN apk add --no-cache bash gcompat libseccomp jq binutils \
 
 # Docker-specific variables
 ENV OASIS_CORE_VERSION=23.0.9
-ENV OASIS_CLI_VERSION=0.6.0
 ENV PARATIME_VERSION=0.7.0-testnet
```
CedarMist:
```suggestion
ENV PARATIME_VERSION=0.7.0
```

abukosek: Good idea, I've bumped the version now.

PR #502, opened 2024-01-05
feat: implement eth_feeHistory endpoint

### LINKED ISSUE #501: eth_feeHistory not supported
The `eth_feeHistory` and `eth_maxPriorityFeePerGas` JSON-RPC calls are used by Web3.py when estimating gas

ptrus:
Fixes https://github.com/oasisprotocol/oasis-web3-gateway/issues/501

Implements eth_feeHistory endpoint. The query supports returning the history for the maximum last 20 (configurable) blocks (this arbitrarily set limit is allowed as per spec: [Requested range of blocks. Clients will return less than the requested range if not all blocks are available.](https://ethereum.github.io/execution-apis/api-documentation/))

To avoid fetching multiple blocks, transactions, and receipts from DB on every fee_history query (which can be expensive) the fee_history for last 20 (configurable) blocks is continuously pre-computed in the gas oracle (only the percentiles are computed on the fly).

TODO:
- [x] test

### DIFF
--- a/gas/backend_test.go
+++ b/gas/backend_test.go
@@ -126,6 +126,11 @@ func TestGasPriceOracle(t *testing.T) {
 	// Default gas price should be returned by the oracle.
 	require.EqualValues(defaultGasPrice.ToBigInt(), gasPriceOracle.GasPrice(), "oracle should return default gas price")
 
+	fh := gasPriceOracle.FeeHistory(10, 10, []float64{0.25, 0.5})
+	require.EqualValues(0, fh.OldestBlock.ToInt().Int64(), "fee history should be empty")
+	require.Empty(0, fh.GasUsedRatio, "fee history should be empty")
+	require.Empty(0, fh.Reward, "fee history should be empty")
+
 	// Emit a non-full block.
 	emitBlock(&emitter, false, nil)
 
@@ -136,7 +141,7 @@ func TestGasPriceOracle(t *testing.T) {
 	emitBlock(&emitter, true, quantity.NewFromUint64(1_000_000_000_000)) // 1000 gwei.
 
 	// 1001 gwei should be returned.
-	require.EqualValues(big.NewInt(1_001_000_000_000), gasPriceOracle.GasPrice(), "oracle should return correct gas price")
+	require.EqualValues(big.NewInt(1_001_000_000_000), gasPriceOracle.GasPrice().ToInt(), "oracle should return correct gas price")
 
 	// Emit a non-full block.
 	emitBlock(&emitter, false, nil)
--- a/gas/fee_history.go
+++ b/gas/fee_history.go
@@ -1,0 +1,181 @@
+package gas
+
+import (
+	"math/big"
+	"slices"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	ethMath "github.com/ethereum/go-ethereum/common/math"
+	"github.com/ethereum/go-ethereum/rpc"
+
+	"github.com/oasisprotocol/oasis-web3-gateway/db/model"
+)
+
+// defaultFeeHistorySize is the default number of recent blocks to store for the fee history query.
+const defaultFeeHistorySize uint64 = 20
+
+// FeeHistoryResult is the result of a fee history query.
+type FeeHistoryResult struct {
+	OldestBlock  *hexutil.Big     `json:"oldestBlock"`
+	Reward       [][]*hexutil.Big `json:"reward,omitempty"`
+	BaseFee      []*hexutil.Big   `json:"baseFeePerGas,omitempty"`
+	GasUsedRatio []float64        `json:"gasUsedRatio"`
+}
+
+type txGasAndReward struct {
+	gasUsed uint64
+	reward  *hexutil.Big
+}
+
+type feeHistoryBlockData struct {
+	height       uint64
+	baseFee      *hexutil.Big
+	gasUsedRatio float64
+	gasUsed      uint64
+
+	// Sorted list of transaction by reward. This is used to
+	// compute the queried percentiles in the fee history query.
+	sortedTxs []*txGasAndReward
+}
+
+// rewards computes the queried reward percentiles for the given block data.
+func (d *feeHistoryBlockData) rewards(percentiles []float64) []*hexutil.Big {
+	rewards := make([]*hexutil.Big, len(percentiles))
+
+	// No percentiles requested.
+	if len(percentiles) == 0 {
+		return rewards
+	}
+
+	// No transactions in the block.
+	if len(d.sortedTxs) == 0 {
+		for i := range rewards {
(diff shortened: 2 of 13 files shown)


CedarMist:
Metamask uses eth_feeHistory to calculate the different transaction priorities for gas price.

See metamask source:
 * [calculateEstimatesForPriorityLevel](https://github.com/MetaMask/core/blob/959b92000d23f46a3c5897b3d14604fdca9906b7/packages/gas-fee-controller/src/fetchGasEstimatesViaEthFeeHistory/calculateGasFeeEstimatesForPriorityLevels.ts#L56)
 * [fetchBlockFeeHistory](https://github.com/MetaMask/core/blob/main/packages/gas-fee-controller/src/fetchBlockFeeHistory.ts#L136)

This doesn't include the current txpool as BTC fee estimation does.

Recently we had a situation where many transactions were queued up, and we were not sure if the fee estimation was returning accurate data, or if the 'Speed Up' button in MetaMask would work correctly.

If `eth_feeHistory` isn't supported it will fallbcak to `eth_gasPrice`, in an ideal situation would be to return the average of the minimum and median for having the transaction included within the next block.

tests/rpc/rpc_test.go
```diff
@@ -146,6 +146,26 @@ func TestEth_GasPrice(t *testing.T) {
 	t.Logf("gas price: %v", price)
 }
 
+func TestEth_FeeHistory(t *testing.T) {
```
ZigaMr, line 158: Just a suggestion, maybe we can test with additional arguments here.
For example 		
`
{
	name: "Basic query with percentiles",
	blockCount: 10,
	lastBlock:  nil,
	percentiles:    []float64{0.25, 0.5, 0.75, 1},
	expectErr:  false,
},
`
`
{
	name: "Query with no reward percentiles",
	blockCount: 5,
	lastBlock:  nil,
	percentiles:    nil,
	expectErr:  false,
},
`
`
{
	name: "Query specific block range",
	blockCount: 3,
	lastBlock:  big.NewInt(5),
	percentiles:    []float64{0.5},
	expectErr:  false,
},
`
`
{
	name: "Invalid block count",
	blockCount: 0,
	lastBlock:  nil,
	percentiles:    []float64{0.5},
	expectErr:  true,
},
`
`
{
	name: "Invalid percentile",
	blockCount: 5,
	lastBlock:  nil,
	percentiles:    []float64{1.5}, // Invalid percentile > 1
	expectErr:  true,
},
`

ZigaMr (approved):
Looks good.

gas/backend_test.go
```diff
@@ -185,4 +190,6 @@ func TestGasPriceOracle(t *testing.T) {
 
 	require.EqualValues(coreClient.minGasPrice.ToBigInt(), gasPriceOracle.GasPrice(), "oracle should return gas reported by the node query")
 	gasPriceOracle.Stop()
+
+	// TODO: emit blocks with transactions, to test fee history.
```
matevz: Is there a difference testing it here compared to rpc_test?

gas/fee_history.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	ethMath "github.com/ethereum/go-ethereum/common/math"
+	"github.com/ethereum/go-ethereum/rpc"
+
+	"github.com/oasisprotocol/oasis-web3-gateway/db/model"
+)
+
+// feeHistoryWindowSize is the number of recent blocks to store for the fee history query.
+const feeHistoryWindowSize = 20
```
matevz: Is this configurable? How did you come up with 20?

gas/fee_history.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// No percentiles requested.
+	if len(percentiles) == 0 {
+		return rewards
+	}
+
+	// No transactions in the block.
+	if len(d.sortedTxs) == 0 {
+		for i := range rewards {
+			rewards[i] = (*hexutil.Big)(common.Big0)
```
matevz, line 54: Should this be the minimum required gas price for Sapphire (0.0000001 per gas unit) instead of 0? I'm not sure if there's another EVM call to fetch that info and dApps will fail otherwise on fresh chains.

tests/rpc/rpc_test.go
```diff
@@ -146,6 +146,26 @@ func TestEth_GasPrice(t *testing.T) {
...
+	for i := 0; i < 5; i++ {
+		receipt := submitTestTransaction(ctx, t)
+		require.EqualValues(t, 1, receipt.Status)
+		require.NotNil(t, receipt)
+	}
+
+	// Query fee history.
+	feeHistory, err := ec.FeeHistory(context.Background(), 10, nil, []float64{0.25, 0.5, 0.75, 1})
+	require.NoError(t, err, "get fee history")
+
```
matevz, line 174: Check that the fees were not zero. (the minimum gas price is set to 0.0000001 in the test docker image, so the fees should also be non-zero)

gas/fee_history.go
ptrus: Arbitrarily chosen, mostly because computing the response is not the most effective, as you must go through all requested blocks. I can make it configurable, but I would keep the default low, unless there is a use case for a larger default.

Most API's I have looked at have a limit of 1024, but they always mention that the node can return less than the requested range if not all blocks are available. The reasonable usages of `eth_feeHistory` for gas estimation usually use small values:
- [web3-py uses 10 blocks for gas estimation](https://github.com/ethereum/web3.py/blob/dc46ee34cc4e3ffa2fa61d223b0d7f13c7e30aaa/web3/_utils/fee_utils.py#L18)
- [multichain uses 10 blocks for gas estimation](https://github.com/renproject/multichain/blob/d2e2ed34004764a5d5ccdae25b4af18a580e4c1c/chain/ethereum/gas.go#L14C2-L14C18)
- Infura used to have a note that only 4 blocks are guaranteed to be available (although, as I see, this note is now gone) and they support up to 1024 blocks (with the note that it could return less)

gas/fee_history.go
ptrus: This is covered by `feehistory.BaseFee` (not via the rewards field), and is handled correctly even if there's no transactions in recent blocks.

tests/rpc/rpc_test.go
ptrus: Added a check to ensure  the returned`feeHistory.BaseFee`'s are > 0.

tests/rpc/rpc_test.go
ptrus: Thanks for the suggestion, added!

gas/backend_test.go
ptrus: This was a lettover comment, removed

gas/fee_history.go
ptrus: This is now also configurable via config.

PR #507, opened 2024-01-15
docker: Rename emerald and sapphire -dev to -localnet

matevz:
We use `-dev` suffix to name the docker images that provide the **compilation** environment for the oasis-core, the key manager and similar. The `-dev` suffix for emerald and sapphire was picked before we started using the *Localnet* term more broadly, specific chain IDs etc. This PR renames the emerald and sapphire docker images running Localnet from `emerald-dev` and `sapphire-dev` to `emerald-localnet` and `sapphire-localnet` respectively to better reflect the actual usage.

### DIFF
--- a/.github/workflows/docker-localnet.yml
+++ b/.github/workflows/docker-localnet.yml
@@ -1,4 +1,4 @@
-name: docker-dev
+name: docker-localnet
 
 on:
   push:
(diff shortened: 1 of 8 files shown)


ptrus (approved):
Fine with me

PR #508, opened 2024-01-15
feat: Bump Sapphire paratime version to 0.7.1-testnet

CedarMist:
sapphire-dev local (oasis-core: 23.0.9, sapphire-paratime: 0.7.1-testnet, oasis-web3-gateway: 4.0.2-git07081eb+dirty)

Noticed some errors when deploying contracts immediately after the container has started, they seem to resolve after a minute or two, trying to track down the cause.

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -109,7 +109,7 @@ jobs:
       OASIS_CORE_VERSION: "23.0.9"
       OASIS_NODE: ${{ github.workspace }}/oasis_core/oasis-node
       OASIS_NET_RUNNER: ${{ github.workspace }}/oasis_core/oasis-net-runner
-      SAPPHIRE_PARATIME_VERSION: 0.7.0-testnet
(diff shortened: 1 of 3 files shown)


kostko (commented):
Should we also bump all of the CI tests in this PR?
https://github.com/oasisprotocol/oasis-web3-gateway/blob/cf1908829e72233509070f6da208b2591b1a98e4/.github/workflows/ci-test.yaml#L112
And possibly also update the expected block gas limit in tests (it is now the same as Mainnet for more consistency).
https://github.com/oasisprotocol/oasis-web3-gateway/blob/cf1908829e72233509070f6da208b2591b1a98e4/tests/rpc/rpc_test.go#L178-L181

CedarMist:
Good catch, I've updated both of those.

However, it reminded me to review places where the `-localnet` vs `-dev` change for the docker image name could need updating.

ptrus (approved):
Great, thanks!

PR #510, opened 2024-01-22
Healthcheck query publickey 

ptrus:
Healthcheck now also periodically tests querying public keys if the Oasis RPC API is enabled (which tests keymanager node access).

Renames metrics prefix from `oasis_oasis_` to `oasis_` (caused due the emerald-web3-gateway -> oasis_web3_gateway rename).

Fixes the `oasis_` RPC endpoint to only be exposed if configured via `ExposeOasisRPCs`.  Because  the `Public` and `Version` fields in `ethRPC.API` were deprecated and not used,  the endpoint was always exposed.

### DIFF
--- a/rpc/oasis/api.go
+++ b/rpc/oasis/api.go
@@ -13,7 +13,7 @@ import (
 
 var ErrInternalError = errors.New("internal error")
 
-// API is the net_ prefixed set of APIs in the Web3 JSON-RPC spec.
+// API is the oasis_ prefixed set of APIs.
 type API interface {
 	// CallDataPublicKey returns the calldata public key for the runtime with the provided ID.
 	CallDataPublicKey(ctx context.Context) (*CallDataPublicKey, error)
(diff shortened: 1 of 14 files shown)


rpc/oasis/api.go
```diff
@@ -9,11 +9,12 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/client"
 	"github.com/oasisprotocol/oasis-sdk/client-sdk/go/modules/core"
+	"github.com/oasisprotocol/oasis-web3-gateway/server"
 )
 
 var ErrInternalError = errors.New("internal error")
 
-// API is the net_ prefixed set of APIs in the Web3 JSON-RPC spec.
+// API is the oasis_ prefixed set of APIs in the Web3 JSON-RPC spec.
```
kostko:
```suggestion
// API is the oasis_ prefixed set of APIs.
```

kostko (approved):

PR #519, opened 2024-02-05
fix(gas): gas price oracle improvements

ptrus:
Gas price oracle heuristics are updated to return better gas price estimates for the `eth_gasPrice` queries. Instead of returning the minimum gas price from the recent blocks, the maximum of the median gas price from recent full blocks is returned. Additionally the gas oracle parameters are now configurable, and the defaults are lowered to better detect full blocks.

### DIFF
--- a/gas/backend.go
+++ b/gas/backend.go
@@ -3,7 +3,6 @@ package gas
 
 import (
 	"context"
-	"math"
 	"sync"
 	"time"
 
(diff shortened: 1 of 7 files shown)


gas/backend.go
```diff
@@ -205,63 +219,61 @@ func (g *gasPriceOracle) indexedBlockWatcher() {
...
-
-	if lastTxPrice == nil {
-		g.Logger.Error("no last tx gas price for block", "block", b)
+	if medTxPrice == nil {
+		g.Logger.Error("no med tx gas price for block", "block", b)
 		return
 	}
-	g.trackPrice(lastTxPrice)
+
+	g.trackPrice(medTxPrice)
```
kostko: Should an epsilon be added here still? Otherwise if the queue is full of transactions with gas price X and the median gas price is also X, your transaction is not likely to make it into the next block? Ideally one would look into the current txqueue as well (to estimate where your transaction would be), but until that is not there, it would be better to still add a small epsilon as long as the blocks are full?

kostko (commented):
Nice, thanks!

gas/backend.go
ptrus: Yeah makes sense :+1:

ptrus: Updated

kostko (approved):

PR #530, opened 2024-02-23
fix: sapphire-localnet: Wait for postgres and the KM to be ready

### LINKED ISSUE #528: docker: Wait for postgres before starting oasis-web3-gateway
Occasionally it happens that oasis-web3-gateway is launched too early and postgres isn't up and running yet. Then `localhost:8545` simply yields `connection refused` for end users.

```
cat /var/log/oasis-web3-gateway.log 
{"addr":"unix:/serverdir/node/net-runner/network/client-0/internal.sock","caller":"main.go:196","level":"info","module":"main","msg":"connecting to local node","ts":"2024-02-22T16:20:06.344504604Z"}
failed to migrate DB: init migrations: dial tcp 127.0.0.1:5432: connect: connection refused
```

Wait for postgres to start up before running oasis-web3-gateay.

abukosek:
Postgres sometimes needs a bit more time to start up, which makes the gateway fail.  We now wait for it before starting the gateway.

Also, the key manager sometimes needs longer to generate the public ephemeral key needed for confidential contracts to work.  We now wait for the epehemeral key to become available before continuing with the rest of the container initialization.

Fixes #528.

### DIFF
--- a/docker/common/start.sh
+++ b/docker/common/start.sh
@@ -74,12 +74,18 @@ notice "Starting oasis-net-runner with ${CYAN}${PARATIME_NAME}${OFF}...\n"
 /spinup-oasis-stack.sh --log.level ${OASIS_NODE_LOG_LEVEL} 2>1 &>/var/log/spinup-oasis-stack.log &
 OASIS_NODE_PID=$!
 
-notice "Starting postgres...\n"
+notice "Waiting for Postgres to start"
 su -c "POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres /usr/local/bin/docker-entrypoint.sh postgres &" postgres 2>1 &>/var/log/postgres.log
+function is_postgres_ready() {
+	pg_isready -h 127.0.0.1 -p 5432 2>1 &>/dev/null
+}
+until is_postgres_ready; do echo -n .; sleep 1; done
+echo
 
-notice "Waiting for Oasis node to start...\n"
+notice "Waiting for Oasis node to start..."
 # Wait for oasis-node socket before starting web3 gateway.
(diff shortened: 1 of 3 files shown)


docker/common/start.sh
```diff
@@ -74,12 +74,18 @@ notice "Starting oasis-net-runner with ${CYAN}${PARATIME_NAME}${OFF}...\n"
 /spinup-oasis-stack.sh --log.level ${OASIS_NODE_LOG_LEVEL} 2>1 &>/var/log/spinup-oasis-stack.log &
 OASIS_NODE_PID=$!
 
-notice "Starting postgres...\n"
+notice "Starting postgres"
```
ptrus: Maybe reword it to `"Waiting for Postgres to start..."` to make it consistent with other notices.

ptrus (approved):
Thanks, this looks really useful!

docker/common/start.sh
abukosek: Thanks, I've done that now :)
The dots are added by the waiting, so the output looks like this most of the time:
```
 * Starting oasis-net-runner with sapphire...
 * Waiting for Postgres to start....
 * Waiting for Oasis node to start...
 * Starting oasis-web3-gateway...
 * Bootstrapping network (this might take a minute)...
 * Waiting for key manager...
 * Populating accounts...
```

ptrus: Makes sense :+1: 

Should the hardcoded three dots (`...`) be removed in other "Waiting for..." lines then as well?

abukosek: No, because in some cases it doesn't have to wait at all and then the output looks weird without the dots :D

matevz:
Thanks, I can confirm this fixed the issue with postgres and the KM as well!

PR #540, opened 2024-03-05
fix: Log transaction data in Debug mode

matevz:
Since the Debug mode is the most verbose one and we typically use it when testing the sapphire-paratime middleware, it makes sense to log the transaction content as well, similar to how we already log the estimate gas and contract calls.


ptrus (approved):

rpc/eth/api.go
```diff
@@ -416,7 +416,7 @@ func (api *publicAPI) Call(ctx context.Context, args utils.TransactionArgs, bloc
 
 func (api *publicAPI) SendRawTransaction(ctx context.Context, data hexutil.Bytes) (common.Hash, error) {
 	logger := api.Logger.With("method", "eth_sendRawTransaction")
-	logger.Debug("request", "length", len(data))
+	logger.Debug("request", "length", len(data), "tx", data)
```
ptrus: maybe remove the `length` now?

PR #544, opened 2024-03-25
feat: allow empty `to` field in `eth_call` to simulate contract deploy

### LINKED ISSUE #543: Hardhat-ignition fails to deploy contracts
It fails with the following error:

`ProviderError: to address not specified`

In `rpc/eth/api.go`  it checks specifically for whether `to` parameter is passed when simulating a transaction, however this means that contract deployments can't be simulated in the way that hardhat-ignition expects.

See relevant code in hardhat-ignition: https://github.com/NomicFoundation/hardhat-ignition/blob/6a2bf9fba8cc2e4c64f8a65efbf8051f43e2fa0e/packages/core/src/internal/execution/future-processor/helpers/network-interaction-execution.ts#L167

```
    at HttpProvider.request (/src/demo-e2eproxy/node_modules/.pnpm/hardhat@2.22.2_ts-node@10.9.2_typescript@5.4.3/node_modules/hardhat/src/internal/core/providers/http.ts:90:21)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async EIP1193JsonRpcClient.call (/src/demo-e2eproxy/node_modules/.pnpm/@nomicfoundation+ignition-core@0.15.0/node_modules/@nomicfoundation/ignition-core/src/internal/execution/jsonrpc-client.ts:321:24)
    at async sendTransactionForOnchainInteraction (/src/demo-e2eproxy/node_modules/.pnpm/@nomicfoundation+ignition-core@0.15.0/node_modules/@nomicfoundation/ignition-core/src/internal/execution/future-processor/helpers/network-interaction-execution.ts:167:28)
    at async sendTransaction (/src/demo-e2eproxy/node_modules/.pnpm/@nomicfoundation+ignition-core@0.15.0/node_modules/@nomicfoundation/ignition-core/src/internal/execution/future-processor/handlers/send-transaction.ts:86:18)
    at async FutureProcessor.processFuture (/src/demo-e2eproxy/node_modules/.pnpm/@nomicfoundation+ignition-core@0.15.0/node_modules/@nomicfoundation/ignition-core/src/internal/execution/future-processor/future-processor.ts:114:9)
    at async ExecutionEngine._executeBatch (/src/demo-e2eproxy/node_modules/.pnpm/@nomicfoundation+ignition-core@0.15.0/node_modules/@nomicfoundation/ignition-core/src/internal/execution/execution-engine.ts:153:30)
    at async ExecutionEngine.executeModule (/src/demo-e2eproxy/node_modules/.pnpm/@nomicfoundation+ignition-core@0.15.0/node_modules/@nomicfoundation/ignition-core/src/internal/execution/execution-engine.ts:114:25)
    at async Deployer.deploy (/src/demo-e2eproxy/node_modules/.pnpm/@nomicfoundation+ignition-core@0.15.0/node_modules/@nomicfoundation/ignition-core/src/internal/deployer.ts:194:25)
    at async SimpleTaskDefinition.action (/src/demo-e2eproxy/node_modules/.pnpm/@nomicfoundation+hardhat-ignition@0.15.0_@nomicfoundation+hardhat-verify@2.0.5_hardhat@2.22.2/node_modules/@nomicfoundation/hardhat-ignition/src/index.ts:220:24)
```

CedarMist:
Closes #543

### DIFF
--- a/rpc/eth/api.go
+++ b/rpc/eth/api.go
@@ -371,7 +371,8 @@ func (api *publicAPI) Call(ctx context.Context, args utils.TransactionArgs, bloc
 	)
 
 	if args.To == nil {
-		return []byte{}, errors.New("to address not specified")
+		// When simulating a contract deploy the To address may not be specified
+		args.To = &common.Address{}
 	}
 	if args.GasPrice != nil {
 		gasPrice = args.GasPrice.ToInt().Bytes()
@@ -396,14 +397,14 @@ func (api *publicAPI) Call(ctx context.Context, args utils.TransactionArgs, bloc
 	}
 
 	res, err := evm.NewV1(api.client).SimulateCall(
-		ctx,
-		round,
-		gasPrice,
-		gas,
-		sender.Bytes(),
-		args.To.Bytes(),
-		amount,
-		input,
+		ctx,             // context
+		round,           // round
+		gasPrice,        // gasPrice
+		gas,             // gasLimit
+		sender.Bytes(),  // caller
+		args.To.Bytes(), // address
+		amount,          // value
+		input,           // data
 	)
 	if err != nil {
 		return nil, api.handleCallFailure(ctx, logger, err)


kostko (approved):

ptrus (approved):
Maybe prefix the commit with `fix: ` so that this ends up as a bugfix in the changelog

rpc/eth/api.go
```diff
@@ -371,7 +371,8 @@ func (api *publicAPI) Call(ctx context.Context, args utils.TransactionArgs, bloc
 	)
 
 	if args.To == nil {
-		return []byte{}, errors.New("to address not specified")
+		// When simulating a contract deploy the To address may not be specified
+		args.To = &common.Address{}
```
ptrus, line 375: So this sets the addres to `0x0000000000000000000000000000000000000000`, i think this should be `nil`. Does the runtime fail with `nil` to address? Should the runtime handle it?

rpc/eth/api.go
```diff
@@ -371,7 +371,8 @@ func (api *publicAPI) Call(ctx context.Context, args utils.TransactionArgs, bloc
 	)
 
 	if args.To == nil {
-		return []byte{}, errors.New("to address not specified")
+		// When simulating a contract deploy the To address may not be specified
```
ptrus, line 374: Also, this should only be allowed if `len(args.data()) != 0` (this is a contract deploy).

rpc/eth/api.go
CedarMist: When I set `To` to `nil` I get:

      error: { code: -32603, message: 'method handler crashed' }

Whereas setting it to all zeros the call succeeds.

I'm having another look at `eth_call` in runtime SDK, it's looking like we might need a separate codepath to handle creation.

ptrus: I think it should be fine to set it to `0x000...` here for now, since fixing (and deploying) it in runtime will take some time in any case 👍  Maybe just open a clickup issue to check if this should be addressed in oasis-sdk.

CedarMist:
This workaround is not good.

PR #546, opened 2024-03-27
fix: allow empty `to` field in `eth_call` to simulate contract deploy

CedarMist:
Depends on #550

* allows empty `to` field in `eth_call`
* requires sapphire-paratime 0.7.2-testnet or above
* adds tests for errors on contract constructor (only for Sapphire)
* makes local testing easier with bind-mounted `/tmp/eth-runtime-test/` to `/serverdir/node/`
 * changes port & network in `conf/tests.yml` so tests can be run against sapphire-localnet` instance (with above bind-mount)

Bugs found:

* Gas estimation of contract creation is too low, can't `eth_call` same message with the gas estimate provided
* Emerald runtime hasn't had an updated release, so new contract constructor tests will fail on Emerald

Why this PR?
 * The `hardhat-ignition` deterministic deployment framework relies on simulating contract creations using `eth_call`


tests/rpc/tx_test.go
```diff
@@ -71,6 +72,133 @@ func waitTransaction(ctx context.Context, ec *ethclient.Client, txhash common.Ha
...
+
+func skipIfNotSapphire(t *testing.T, ec *ethclient.Client) {
+	chainID, err := ec.ChainID(context.Background())
+	require.Nil(t, err, "get chainid")
+
+	switch chainID.Uint64() {
+	case 0x5afd:
+	case 0x5afe:
+	case 0x5aff:
+		break
```
ptrus: This break is not needed as it is implicit already. Also the cases don't `fallthrough` (unless specified via `falthtrough`).

So can just do 
```
	switch chainID.Uint64() {
	case 0x5afd:
	case 0x5afe:
	case 0x5aff:
        default:
           t.Skip("Skipping contract constructor tests on non-Sapphire chain")
```
or
```
	switch chainID.Uint64() {
	case 0x5afd, 0x5afe, 0x5aff:
        default:
           t.Skip("Skipping contract constructor tests on non-Sapphire chain")

```

docker/common/start.sh
```diff
@@ -87,6 +87,11 @@ notice "Waiting for Oasis node to start..."
 while [[ ! -S ${OASIS_NODE_SOCKET} ]]; do echo -n .; sleep 1; done
 echo
 
+chmod 755 /serverdir/node/net-runner/
```
ptrus, line 92: Please add a comment above the block why this is done (e.g. to fix permissions in case of mounting host volumes)

tests/tools/spinup-oasis-stack.sh
```diff
@@ -19,8 +19,12 @@ function paratime_ver {
 export FIXTURE_FILE="${OASIS_NODE_DATADIR}/fixture.json"
 export STAKING_GENESIS_FILE="$(dirname "$0")/staking_genesis.json"
 
-rm -rf "$OASIS_NODE_DATADIR"
-mkdir -p "$OASIS_NODE_DATADIR"
+rm -rf "$OASIS_NODE_DATADIR/net-runner" || true
+rm -rf "$OASIS_NODE_DATADIR/net-runner.log" || true
+rm -rf "$OASIS_NODE_DATADIR/fixture.json" || true
+# When $OASIS_NODE_DATADIR is bind-mounted, below fails, but the above succeed
```
ptrus, line 25: Can probably remove the `|| true` part in the above ones.

ptrus:
> makes local testing easier with bind-mounted /tmp/eth-runtime-test/ to /serverdir/node/

Can you add a short note somewhere, maybe into README, on the workflow for local testing with this.

ptrus (approved):
Thanks, very nice!

PR #593, opened 2024-07-09
fix: Fix sapphire-localnet startup on Mac Docker

abukosek:
This PR fixes `sapphire-localnet` startup in Docker on Mac, as reported by @lubej.

### DIFF
--- a/docker/common/start.sh
+++ b/docker/common/start.sh
@@ -11,8 +11,6 @@
 # - BEACON_BACKEND: beacon epoch transition mode 'mock' (default) or 'default'
 # - OASIS_SINGLE_COMPUTE_NODE: (default: true) if non-empty only run a single compute node
 
-set -euo pipefail
-
 export OASIS_DOCKER_NO_GATEWAY=${OASIS_DOCKER_NO_GATEWAY:-no}
 
 export OASIS_NODE_LOG_LEVEL=${OASIS_NODE_LOG_LEVEL:-warn}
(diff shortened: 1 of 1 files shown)


lubej:
Closes https://github.com/oasisprotocol/oasis-web3-gateway/issues/592

kostko (commented):
Does something related to version determination fail? What is the "version line" output on the failing machine?

lubej:
> Does something related to version determination fail? What is the "version line" output on the failing machine?

image

kostko:
Thanks, but can you also show the exact output of the "version line" after you run the Docker image with the proposed workaround from this PR.

lubej:
> Thanks, but can you also show the exact output of the "version line" after you run the Docker image with the proposed workaround from this PR.

There was no version line, it failed beforehand.

kostko:
As mentioned in the comment, what is the output with the proposed fix from this PR.

lubej:
> As mentioned in the comment, what is the output with the proposed fix from this PR.

Here is the output(Docker image build by @abukosek, and tested on my machine):

image

kostko (approved):

abukosek:
Here is what I think was the problem and why just moving `set -euo pipefail` a few lines down is a good enough solution.

The initial problem was that when running the image on Docker Mac on a machine with Apple M silicon (with the appropriate `--platform linux/x86_64` flag), the startup script (`/start.sh`) immediately exited with error code 141, which means that it received a SIGPIPE.

We narrowed this down to the following line in the script:
```
OASIS_WEB3_GATEWAY_VERSION=$(${OASIS_WEB3_GATEWAY} -v | head -n1 | cut -d " " -f 3 | sed -r 's/^v//')
```
Sometimes it looked like `head` failed or `cut` failed or even `sed`, but it was all random.
It makes no sense that these commands would fail, since the Docker image is the same as on Linux and there it works perfectly.

However, on Apple M silicon, Docker images run in an emulation layer and run more slowly.

Normally, this wouldn't be a problem, but in this case, the output from `oasis-web3-gateway -v` has multiple lines, but with `head -n1`, we only take the first one.
If the emulation is slow, the `oasis-web3-gateway` might still be outputting the version information, while `head -n1` has already taken the first line **and closed the pipe**.  Closing the pipe while the writing end isn't done yet results in a broken pipe.
Since we have `pipefail` enabled, this causes the entire startup script to fail with error 141.

This also explains why the error was so random -- sometimes the process managed to output everything on time and exited cleanly, but sometimes it was interrupted by the broken pipe and brought down the entire script with it.

So, moving the `set -euo pipefail` command below these initial lines that extract the version etc. is a good enough solution, since we only care about the first line of the version output and don't care if the process isn't done writing the rest of the version information, as we don't need it anyway.

PR #608, opened 2024-09-05
fix: Add script that waits until container startup is done

abukosek:
This PR adds a `/wait-container-ready.sh` script to the `sapphire-localnet` and `emerald-localnet` Docker images, so that external tools and processes can just run that and it will wait until the local network has fully finished starting up and all test accounts are populated.

### DIFF
--- a/docker/common/start.sh
+++ b/docker/common/start.sh
@@ -11,6 +11,8 @@
 # - BEACON_BACKEND: beacon epoch transition mode 'mock' (default) or 'default'
 # - OASIS_SINGLE_COMPUTE_NODE: (default: true) if non-empty only run a single compute node
 
+rm -f /CONTAINER_READY
+
 export OASIS_DOCKER_NO_GATEWAY=${OASIS_DOCKER_NO_GATEWAY:-no}
 
 export OASIS_NODE_LOG_LEVEL=${OASIS_NODE_LOG_LEVEL:-warn}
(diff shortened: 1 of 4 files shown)


docker/common/start.sh
```diff
@@ -168,7 +170,7 @@ if [[ $OASIS_DOCKER_NO_GATEWAY != 'yes' && $PARATIME_NAME == 'sapphire' ]]; then
           http://127.0.0.1:8545 2>1 | jq -e '.result | has("key")' 2>1 &>/dev/null
     }
     until is_km_ready; do
-        if [[ ${BEACON_BACKEND} == 'mock' ]]; then
+        if [[ "${BEACON_BACKEND}" == "mock" ]]; then
```
CedarMist, line 173: Please note, with double bracketed if, it's safe to use just `if [[ ${BLAH} == "yes" ]]` or `if [[ "${BLAH}" == "yes" ]]`

docker/common/start.sh
```diff
@@ -159,7 +161,7 @@ else
 fi
 echo
 
-if [[ $OASIS_DOCKER_NO_GATEWAY != 'yes' && $PARATIME_NAME == 'sapphire' ]]; then
+if [[ "x${OASIS_DOCKER_NO_GATEWAY}" != "xyes" && "${PARATIME_NAME}" == "sapphire" ]]; then
```
CedarMist, line 164: `"x$VAR" != "xyes"` is archaic syntax for non-bash systems using the traditional `[` command, it's not necessary with native `if [[` and `${BRACKET}` parameter substitution, but `"${BRACKET}"` is still good practice. 

Double brackets `[[ ... ]]` prevent word splitting and globbing on the variable expansion etc.

CedarMist (approved):

docker/common/start.sh
abukosek: I just modified this to make it consistent with the style used in the rest of the script and in case we decide later to switch from bash to the standard sh that's in the base Alpine image, as someone suggested a while ago :)

PR #618, opened 2024-09-17
feat: detect & log signed queries via eth_call

### LINKED ISSUE #604: Track how many times signed queries are being used as a metric for graphana
A counter for the number of times `eth_call` is used with signed queries.

To detect if this feature is actively being used.

CedarMist:
Fixes: https://github.com/oasisprotocol/oasis-web3-gateway/issues/604

Adds `oasis_web3_gateway_signed_queries` prometheus counter.

### DIFF
--- a/rpc/eth/metrics/api.go
+++ b/rpc/eth/metrics/api.go
@@ -13,7 +13,9 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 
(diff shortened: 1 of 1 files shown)


ptrus (approved):

ptrus:
Do we want to get this merged?

matevz (approved):
I wonder how significant is the performance hit of the extra unmarshalling.

ptrus:
> I wonder how significant is the performance hit of the extra unmarshalling.

We talked and think it should be negligible. We also don't expect a lot of signed queries.

PR #623, opened 2024-09-20
docker: Fallback sapphire-localnet to glibc

matevz:
Workaround for https://github.com/oasisprotocol/oasis-web3-gateway/issues/622


jberci (approved):
I didn't manage to reproduce the original issue, but the changes produce a usable docker image on an M1 computer.

PR #625, opened 2024-09-22
fix: Enable expensive txes for Localnet compute node

matevz:
This PR enables `allow_expensive_queries` parameter on compute-0 node too, since it's required for ROFL queries.


matevz:
Perhaps we could/should run ROFL on the client-0 node instead of compute-0 and this would not be needed.

ptrus (approved):

matevz:
> Perhaps we could/should run ROFL on the client-0 node instead of compute-0 and this would not be needed.

Running ROFLs that are part of the ParaTime .orc is currently disabled on client, so this is currently not straight forward to do. Maybe sth for the future, when we revamp the test-runner and net-runner code.

PR #626, opened 2024-09-26
docker/sapphire: Include Envoy proxy

### LINKED ISSUE #423: sapphire-localnet: expose internal.sock
This will allow Go client and Oasis CLI can connect from outside docker.

One option would be to do a bind-mount to expose the `client-0` directory:

    OASIS_NODE_SOCKET=${OASIS_NODE_DATADIR}/net-runner/network/client-0/internal.sock

Another cleaner option would be to expose over TCP.

Or, may be able to link two containers together and directly expose the socket from one to another?

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-web3-gateway/issues/423

Envoy proxy is now included which exposes the clients node `internal.sock` on port `8544` (by default). Envoy is used because it also supports grpc-web, which might be useful/needed when testing any dapps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ the full gateway & paratime stack together.
 For running tests, start the docker network without the gateway, by setting the `OASIS_DOCKER_NO_GATEWAY=yes` environment variable:
 
 ```bash
-docker run --rm -ti -e OASIS_DOCKER_NO_GATEWAY=yes -p5432:5432 -p8545:8545 -p8546:8546 -v /tmp/eth-runtime-test:/serverdir/node ghcr.io/oasisprotocol/sapphire-localnet:local -test-mnemonic -n 4
+docker run --rm -ti -e OASIS_DOCKER_NO_GATEWAY=yes -p5432:5432 -p8544-8546:8544-8546 -v /tmp/eth-runtime-test:/serverdir/node ghcr.io/oasisprotocol/sapphire-localnet:local -test-mnemonic -n 4
 ```
 
 Once bootstrapped, run the tests:
--- a/docker/common/start.sh
+++ b/docker/common/start.sh
@@ -11,6 +11,8 @@
 # - BEACON_BACKEND: beacon epoch transition mode 'mock' (default) or 'default'
 # - OASIS_SINGLE_COMPUTE_NODE (default: true): if non-empty only run a single compute node
 # - OASIS_CLI_BINARY (optional): path to oasis binary. If provided, Oasis CLI will be configured for Localnet
+# - ENVOY_BINARY (optional): path to Envoy binary. If provided, Envoy proxy to Oasis Client node will be started
+# - ENVOY_CONFIG_FILE: path to Envoy config file. Required if ENVOY_BINARY is provided
 
 rm -f /CONTAINER_READY
 
@@ -38,6 +40,7 @@ OASIS_KM_SOCKET=${OASIS_NODE_DATADIR}/net-runner/network/keymanager-0/internal.s
 
 OASIS_WEB3_GATEWAY_PID=""
 OASIS_NODE_PID=""
+ENVOY_PID=""
 
 set -euo pipefail
 
@@ -47,6 +50,9 @@ function cleanup {
   fi
   if [[ -n "${OASIS_NODE_PID}" ]]; then
     kill -9 ${OASIS_NODE_PID}
+  fi
+  if [[ -n "${ENVOY_PID}" ]]; then
+    kill -9 ${ENVOY_PID}
   fi
 }
 
@@ -132,6 +138,17 @@ notice "Waiting for Oasis node to start..."
 # Wait for oasis-node socket before starting web3 gateway.
 while [[ ! -S ${OASIS_NODE_SOCKET} ]]; do echo -n .; sleep 1; done
 echo
+
+if [ ! -z "${ENVOY_BINARY:-}" ]; then
+  notice "Waiting for Envoy proxy to start"
+
+  ${ENVOY_BINARY} -c /envoy.yml 2>1 &>/var/log/envoy.log &
+  ENVOY_PID=$!
+
+  # Wait for Envoy to start.
+  while ! curl -s http://localhost:8544/ 2>1 &>/dev/null; do echo -n .; sleep 1; done
+  echo
+fi
 
 # Fixes permissions so oasis-web3-gateway tests can be run
 # While /serverdir/node is bind-mounted to /tmp/eth-runtime-test
@@ -257,7 +274,8 @@ fi
 
 echo
 printf "${YELLOW}WARNING: The chain is running in ephemeral mode. State will be lost after restart!${OFF}\n\n"
-notice "Listening on ${CYAN}http://localhost:8545${OFF} and ${CYAN}ws://localhost:8546${OFF}. Chain ID: ${GATEWAY__CHAIN_ID}\n"
+notice "GRPC listening on ${CYAN}http://localhost:8544${OFF}.\n"
+notice "Web3 RPC listening on ${CYAN}http://localhost:8545${OFF} and ${CYAN}ws://localhost:8546${OFF}. Chain ID: ${GATEWAY__CHAIN_ID}.\n"
 notice "Container start-up took ${CYAN}$((T_END-T_START))${OFF} seconds, node log level is set to ${CYAN}${OASIS_NODE_LOG_LEVEL}${OFF}.\n"
 
 touch /CONTAINER_READY
--- a/docker/emerald-localnet/Dockerfile
+++ b/docker/emerald-localnet/Dockerfile
@@ -5,9 +5,19 @@ COPY . /go/oasis-web3-gateway
 RUN cd oasis-web3-gateway && make && strip -S -x oasis-web3-gateway docker/common/oasis-deposit/oasis-deposit
 
 # Build emerald-localnet
-FROM postgres:16-alpine
-RUN apk add --no-cache bash gcompat libseccomp jq binutils \
-	&& su -c "POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres /usr/local/bin/docker-entrypoint.sh postgres &" postgres
+FROM postgres:16-bookworm
+RUN apt update && apt install -y bash libseccomp2 unzip jq binutils curl wget && apt clean \
+    && su -c "POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres /usr/local/bin/docker-entrypoint.sh postgres &" postgres \
+    # Install Envoy.
+    && curl -fsSL https://apt.envoyproxy.io/signing.key | gpg --dearmor -o /etc/apt/keyrings/envoy-keyring.gpg \
+    && echo "deb [signed-by=/etc/apt/keyrings/envoy-keyring.gpg] https://apt.envoyproxy.io bookworm main" | tee /etc/apt/sources.list.d/envoy.list \
+    && apt update \
+    && apt install -y envoy \
+    # Clean up.
+    && apt clean \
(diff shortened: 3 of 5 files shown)


docker/common/start.sh
```diff
@@ -133,6 +137,100 @@ notice "Waiting for Oasis node to start..."
...
+        "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
+        explicit_http_config:
+          http2_protocol_options: {}
+layered_runtime:
+  layers:
+  - name: static
+    static_layer:
+      re2:
+        max_program_size:
+          error_level: 1000000
```
matevz: I suggest we move this to a separate file similar to `localnet.yml` for the web3 gateway.

docker/common/start.sh
```diff
@@ -133,6 +137,100 @@ notice "Waiting for Oasis node to start..."
...
+EOF
+
+  ${OASIS_ENVOY_BINARY} -c /envoy.yml 2>1 &>/var/log/envoy.log &
+
+  # Wait for Envoy to start.
+  while ! curl -s http://localhost:${OASIS_ENVOY_PORT}/ 2>1 &>/dev/null; do echo -n .; sleep 1; done
+  echo
```
matevz, line 150: Can you also add something like `* GRPC listening on http://localhost:8544` at the end below the existing RPC. And for the existing RPC change text to sth like `* Web3 RPC listening on http://localhost:8545...`.

docker/sapphire-localnet/Dockerfile
```diff
@@ -100,6 +112,7 @@ ENV OASIS_NODE_BINARY=/oasis-node
 ENV OASIS_NET_RUNNER_BINARY=/oasis-net-runner
 ENV OASIS_NODE_DATADIR=/serverdir/node
 ENV OASIS_WEB3_GATEWAY_BINARY=/oasis-web3-gateway
+ENV OASIS_ENVOY_BINARY=envoy
```
matevz: I would prefer absolute paths just to be sure.

```suggestion
ENV OASIS_ENVOY_BINARY=/usr/bin/envoy
```

matevz (commented):
Can you also please add envoy support to emerald-localnet.

docker/common/start.sh
ptrus: Done

docker/common/start.sh
ptrus: Updated, this is now displayed at the end of startup:

```
WARNING: The chain is running in ephemeral mode. State will be lost after restart!

 * Chain ID: 0x5afd
 * GRPC listening on http:://localhost:8544.
 * Web3 RPC listening on http://localhost:8545 and ws://localhost:8546.
 * Container start-up took 105 seconds, node log level is set to warn.
```

README.md
```diff
@@ -46,7 +46,7 @@ the full gateway & paratime stack together.
 For running tests, start the docker network without the gateway, by setting the `OASIS_DOCKER_NO_GATEWAY=yes` environment variable:
 
 ```bash
-docker run --rm -ti -e OASIS_DOCKER_NO_GATEWAY=yes -p5432:5432 -p8545:8545 -p8546:8546 -v /tmp/eth-runtime-test:/serverdir/node ghcr.io/oasisprotocol/sapphire-localnet:local -test-mnemonic -n 4
+docker run --rm -ti -e OASIS_DOCKER_NO_GATEWAY=yes -p5432:5432 -p8544-8546:8544-8546 -v /tmp/eth-runtime-test:/serverdir/node ghcr.io/oasisprotocol/sapphire-localnet:local -test-mnemonic -n 4
```
matevz, line 49: -test-mnemonic is also enabled by default now.

```suggestion
docker run --rm -ti -e OASIS_DOCKER_NO_GATEWAY=yes -p5432:5432 -p8544-8546:8544-8546 -v /tmp/eth-runtime-test:/serverdir/node ghcr.io/oasisprotocol/sapphire-localnet:local -n 4
```

docker/common/start.sh
```diff
@@ -257,7 +274,9 @@ fi
 
 echo
 printf "${YELLOW}WARNING: The chain is running in ephemeral mode. State will be lost after restart!${OFF}\n\n"
-notice "Listening on ${CYAN}http://localhost:8545${OFF} and ${CYAN}ws://localhost:8546${OFF}. Chain ID: ${GATEWAY__CHAIN_ID}\n"
+notice "Chain ID: ${GATEWAY__CHAIN_ID}\n"
+notice "GRPC listening on ${CYAN}http:://localhost:8544${OFF}.\n"
+notice "Web3 RPC listening on ${CYAN}http://localhost:8545${OFF} and ${CYAN}ws://localhost:8546${OFF}.\n"
```
matevz: Actually the Chain ID is related to the Web3 RPC only. So I would put it at the end of the "Web3 RPC..." line as it was now.

docker/common/start.sh
```diff
@@ -11,6 +11,8 @@
 # - BEACON_BACKEND: beacon epoch transition mode 'mock' (default) or 'default'
 # - OASIS_SINGLE_COMPUTE_NODE (default: true): if non-empty only run a single compute node
 # - OASIS_CLI_BINARY (optional): path to oasis binary. If provided, Oasis CLI will be configured for Localnet
+# - ENVOY_BINARY (optional): path to Envoy binary. If provided, Envoy proxy to Oasis Client node will be started.
+# - ENVOY_CONFIG_FILE: path to Envoy config file. Required if ENVOY_BINARY is provided.
```
matevz:
```suggestion
# - ENVOY_BINARY (optional): path to Envoy binary. If provided, Envoy proxy to Oasis Client node will be started
# - ENVOY_CONFIG_FILE: path to Envoy config file. Required if ENVOY_BINARY is provided
```

docker/common/start.sh
matevz: Commented below.

docker/emerald-localnet/Dockerfile
```diff
@@ -5,9 +5,21 @@ COPY . /go/oasis-web3-gateway
 RUN cd oasis-web3-gateway && make && strip -S -x oasis-web3-gateway docker/common/oasis-deposit/oasis-deposit
 
 # Build emerald-localnet
-FROM postgres:16-alpine
-RUN apk add --no-cache bash gcompat libseccomp jq binutils \
-	&& su -c "POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres /usr/local/bin/docker-entrypoint.sh postgres &" postgres
+FROM postgres:16-bookworm
+RUN apt update && apt install -y bash libseccomp2 unzip jq binutils curl wget && apt clean \
+    && su -c "POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres /usr/local/bin/docker-entrypoint.sh postgres &" postgres \
+
```
matevz: I think docker has issues with empty lines inside the RUN command. I'm getting
```
[WARNING]: Empty continuation line found in: RUN apt update && apt install -y bash libseccomp2 unzip jq binutils curl wget && apt clean     && su -c "POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres /usr/local/bin/docker-entrypoint.sh postgres &" postgres     && curl -fsSL https://apt.envoyproxy.io/signing.key | gpg --dearmor -o /etc/apt/keyrings/envoy-keyring.gpg     && echo "deb [signed-by=/etc/apt/keyrings/envoy-keyring.gpg] https://apt.envoyproxy.io bookworm main" | tee /etc/apt/sources.list.d/envoy.list     && apt update     && apt install -y envoy     && apt clean     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*     && rm -rf /etc/apt/sources.list.d/envoy.list     && rm -f /etc/apt/keyrings/envoy-keyring.gpg
[WARNING]: Empty continuation lines will become errors in a future release.
```

Let's move all the comments before the command.

docker/sapphire-localnet/Dockerfile
```diff
@@ -79,13 +79,21 @@ RUN wget https://github.com/oasisprotocol/${PARATIME_NAME}-paratime/releases/dow
...
-#FROM postgres:16-alpine
-#RUN apk add --no-cache bash gcompat libseccomp jq binutils curl \
-#    && su -c "POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres /usr/local/bin/docker-entrypoint.sh postgres &" postgres
-# Fallback to Debian-based image. See https://github.com/oasisprotocol/oasis-web3-gateway/issues/622
-FROM postgres:16
+FROM postgres:16-bookworm
 RUN apt update && apt install -y bash libseccomp2 unzip jq binutils curl && apt clean \
-    && su -c "POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres /usr/local/bin/docker-entrypoint.sh postgres &" postgres
+    && su -c "POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres /usr/local/bin/docker-entrypoint.sh postgres &" postgres \
+
```
matevz: Remove any empty lines and move the comments before the RUN command.

matevz (commented):
I tried testing it locally with the Oasis CLI, but it seems a TLS connection is required. Do you think we could simply use a self-signed cert?

```
oasis net add localnet a41bd4bc75f31e68ceb1e8543f1a08ca54069b155f980bdf0316ec16ed194772 localhost:8544
oasis net status --network localnet
Error: failed to retrieve remote node's chain context: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
```

docker/emerald-localnet/Dockerfile
ptrus: I removed the empty lines and the warning is gone. I think the comments-only lines should be fine to keep. 

There was a bug in some Docker versions where it incorrectly warned about  comment-only lines as well (https://github.com/moby/moby/pull/35004).

ptrus:
> I tried testing it locally with the Oasis CLI, but it seems a TLS connection is required. Do you think we could simply use a self-signed cert?

Added. The GRPC port now supports both TLS with self-signed certs and unsecured traffic as well.

```
oasis net add localnet 761298f959ad9f7fd8f832cb62fa0c15c2ba9a112a27fabfcb6bd6fca6d47843 localhost:8544
? Description: just testing
? Denomination symbol: TEST
? Denomination decimal places: 9
```

matevz:
Oasis CLI is still not entirely satisfied ;)

```
$ oasis net add localnet 0b91b8e4e44b2003a7c5e23ddadb5e14ef5345c0ebcb3ddcae07fa2f244cab76 localhost:8544
? Description: 
? Denomination symbol: TEST
? Denomination decimal places: 9
$ oasis net status --network localnet
Error: failed to retrieve remote node's chain context: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate relies on legacy Common Name field, use SANs instead"
```

ptrus:
Fixed, but it turns out that the oasis-cli doesn't allow self-signed certs

```
oasis net status --network localnet_test
Error: failed to retrieve remote node's chain context: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate signed by unknown authority"
```

IMO this should be fixed this in the CLI so that it supports untrusted connection for localhost (or via a flag).

ptrus:
Removed TLS support as discused. `oasis-cli` and `oasis-node` will start working once the new releases will be made and binaries will be updated. I suggest we merge this PR, not blocking on that.

matevz (approved):
I tested this with the fork of Oasis CLI using https://github.com/oasisprotocol/oasis-sdk/pull/2013 and works. Good job!

PR #636, opened 2024-10-21
other(localnet): use wait-ready on keymanager instead of polling

ptrus:
Followup to https://github.com/oasisprotocol/oasis-web3-gateway/pull/635


matevz (approved):
Looks good.

PR #637, opened 2024-10-22
feat(localnet): Include Nexus and Explorer in localnet images

### LINKED ISSUE #632: sapphire-localnet: Integrate Oasis Explorer
Currently, debugging dApps is really annoying on Localnet, since you cannot examine contracts and transactions in a nice Explorer-like way. Integrate Nexus and Explorer into sapphire-localnet (and emerald-localnet) and enable it by default on port 80. Add `--no-explorer` do disable those used on CI to save resources.

ptrus:
Closes https://github.com/oasisprotocol/oasis-web3-gateway/issues/632

I think this is ready to be merged. For now, we use my fork of the explorer (with a hacky fix, to support defining enabled paratimes), as soon as some initial support is added in the explorer repo, we can switch to that: https://github.com/oasisprotocol/explorer/issues/1597


docker/common/start.sh
```diff
@@ -226,6 +237,35 @@ else
...
+
+  # Wait for Oasis Nexus to start.
+  while ! curl -s http://localhost:8547/ 2>1 &>/dev/null; do echo -n .; sleep 1; done
+  echo
+fi
+
+if [ ! -z "${OASIS_EXPLORER_DIR:-}" ]; then
+  notice "Waiting for Explorer to start"
+  cd ${OASIS_EXPLORER_DIR}
+  serve -s ${OASIS_EXPLORER_DIR} -l 3000 > /var/log/explorer.log 2>&1 &
```
matevz: The explorer should probably become the default web interface to our Localnet. Also, let''s move this to a constant.

```suggestion
  serve -s ${OASIS_EXPLORER_DIR} -l 80 > /var/log/explorer.log 2>&1 &
```

docker/sapphire-localnet/Dockerfile
```diff
@@ -78,10 +78,54 @@ RUN wget https://github.com/oasisprotocol/${PARATIME_NAME}-paratime/releases/dow
...
+    && NODE_ENV=production yarn build
+
 # Build sapphire-localnet
 FROM postgres:16-bookworm
-RUN apt update && apt install -y bash libseccomp2 unzip jq binutils curl && apt clean \
+RUN apt update && apt install -y bash libseccomp2 unzip jq binutils curl python3 \
     && su -c "POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres /usr/local/bin/docker-entrypoint.sh postgres &" postgres \
+    # Explorer frontend deps.
+    && curl -fsSL https://deb.nodesource.com/setup_20.x | bash - \
+    && apt install -y nodejs && npm install -g serve \
```
matevz, line 128: Do you need nodejs for serving the explorer? I would simply use https://docs.python.org/3/library/http.server.html:
```
python -m http.server 80 --bind 127.0.0.1 --directory /explorer
```

CedarMist:
Awesome!

image

docker/sapphire-localnet/Dockerfile
```diff
@@ -78,10 +77,55 @@ RUN wget https://github.com/oasisprotocol/${PARATIME_NAME}-paratime/releases/dow
...
+# RUN wget https://github.com/oasisprotocol/nexus/releases/download/v${NEXUS_VERSION}/oasis_nexus_${NEXUS_VERSION}_linux_amd64.tar.gz \
+#     && tar -zxvf oasis_nexus_${NEXUS_VERSION}_linux_amd64.tar.gz \
+#     && strip -S -x /oasis_nexus_${NEXUS_VERSION}_linux_amd64/nexus \
+#     && mv /oasis_nexus_${NEXUS_VERSION}_linux_amd64/nexus / \
+#     && rm -rf /oasis_nexus_${NEXUS_VERSION}_linux_amd64* && \
+#     # We also need to fetch the storage migrations.
+#     git clone https://github.com/oasisprotocol/nexus.git nexus-git --branch v${NEXUS_VERSION} --depth 1 \
+#     && mv /nexus-git/storage/migrations /nexus-migrations \
+#     && rm -rf nexus-git
+# TODO: change to release once https://github.com/oasisprotocol/nexus/pull/766 is released.
```
CedarMist: https://github.com/oasisprotocol/nexus/pull/766 is merged :+1: 

Can comment out the alternate below or replace `NEXUS_VERSION=main`?

docker/sapphire-localnet/Dockerfile
```diff
@@ -78,10 +77,55 @@ RUN wget https://github.com/oasisprotocol/${PARATIME_NAME}-paratime/releases/dow
...
+ARG EXPLORER_REPO=https://github.com/ptrus/explorer.git
+ARG EXPLORER_VERSION=1a78782e10fa266707a428c5fd8ba91966d588a1
+COPY docker/common/explorer.env /nexus.env.production
+RUN git clone ${EXPLORER_REPO} explorer --depth 1 \
+    && cd explorer \
+    && git fetch origin ${EXPLORER_VERSION} --depth 1 \
+    && git checkout FETCH_HEAD \
+    && mv /nexus.env.production .env.production \
+    && echo "REACT_APP_LOCALNET_SAPPHIRE=true\nREACT_APP_LOCALNET_EMERALD=false" >> .env.production \
+    && curl -fsSL https://deb.nodesource.com/setup_20.x | bash - \
```
CedarMist, line 116: Can these be added to oasis-core-dev Docker in future?

docker/sapphire-localnet/Dockerfile
```diff
@@ -78,10 +77,55 @@ RUN wget https://github.com/oasisprotocol/${PARATIME_NAME}-paratime/releases/dow
...
+    && NODE_ENV=production yarn build
+
 # Build sapphire-localnet
 FROM postgres:16-bookworm
-RUN apt update && apt install -y bash libseccomp2 unzip jq binutils curl && apt clean \
+RUN apt update && apt install -y bash libseccomp2 unzip jq binutils curl \
     && su -c "POSTGRES_USER=postgres POSTGRES_PASSWORD=postgres /usr/local/bin/docker-entrypoint.sh postgres &" postgres \
+    # Explorer frontend deps.
+    && curl -fsSL https://deb.nodesource.com/setup_20.x | bash - \
+    && apt install -y nodejs && npm install -g serve \
```
CedarMist, line 128: Is it worth merging frontend & envoy dependencies into a single `apt install` command in future to speed-up container build time?

docker/common/start.sh
```diff
@@ -284,6 +326,7 @@ echo
 printf "${YELLOW}WARNING: The chain is running in ephemeral mode. State will be lost after restart!${OFF}\n\n"
 notice "GRPC listening on ${CYAN}http://localhost:8544${OFF}.\n"
 notice "Web3 RPC listening on ${CYAN}http://localhost:8545${OFF} and ${CYAN}ws://localhost:8546${OFF}. Chain ID: ${GATEWAY__CHAIN_ID}.\n"
+notice "Localnet Explorer available at ${CYAN}http://localhost:3000${OFF}. Nexus API at ${CYAN}http://localhost:8547${OFF}\n"
```
CedarMist: Would be better to have the Explorer & Nexus API URLs on separate lines.  I can do this in a follow-on PR quickly.

docker/sapphire-localnet/Dockerfile
ptrus: Yes could be, although here it's easier to track why it is needed.

docker/sapphire-localnet/Dockerfile
ptrus: If we have a strong preference for that, we could. I like it in a separate section since then it's clearer why it's needed and it can easily be removed in case it's not needed anymore in future.

Btw this is currently only needed to serve the static explorer frontend via `nodejs serve`, we could switch to something more lightweight like `nginx`. I  used this for now since it was the fastest to setup and it worked.

docker/common/start.sh
ptrus: Updated

docker/sapphire-localnet/Dockerfile
CedarMist: Hmm, the Docker already has Envoy but it can't serve static files, and doesn't have Python. OpenSSL is installed and has a built-in web server but it only does HTTPS.

I guess this works for now.

matevz (approved):
Did you add the --no-explorer flag?

docker/sapphire-localnet/Dockerfile
```diff
@@ -78,10 +78,54 @@ RUN wget https://github.com/oasisprotocol/${PARATIME_NAME}-paratime/releases/dow
...
+    && echo "REACT_APP_LOCALNET_SAPPHIRE=true\nREACT_APP_LOCALNET_EMERALD=false" >> .env.production \
+    && curl -fsSL https://deb.nodesource.com/setup_20.x | bash - \
+    && apt install -y nodejs \
+    && npm install -g yarn \
+    && yarn install \
+    && NODE_ENV=production yarn build
+
 # Build sapphire-localnet
 FROM postgres:16-bookworm
-RUN apt update && apt install -y bash libseccomp2 unzip jq binutils curl && apt clean \
```
abukosek, line 83: Don't forget to add `apt clean` back to reduce the image size a bit :)

docker/sapphire-localnet/Dockerfile
ptrus: Python is not currently installed in the container, so it would also require adding a dependency.

Also, the default Python HTTP server doesn’t work nicely with reacts router. Refreshing the page while on a subpage results in a 404 error, while `serve` handles this nicely.

docker/sapphire-localnet/Dockerfile
ptrus: We already have one at the end of this RUN command (line 135), so I removed this one.

ptrus:
> Did you add the --no-explorer flag?

If there’s a specific use case for it, we can consider adding it, but I wouldn’t include it just for the sake of having it.

matevz:
> > Did you add the --no-explorer flag?
> 
> If there’s a specific use case for it, we can consider adding it, but I wouldn’t include it just for the sake of having it.

To avoid nexus compute and storage cycles during CI tests for your dApp?

@abukosek also noticed considerable longer startup time (~10s)

PR #641, opened 2024-11-07
feat(localnet): update nexus to 0.4.1 and exlporer

ptrus:
Updates:
- nexus to 0.4.1
- explorer to include https://github.com/oasisprotocol/explorer/pull/1628

Test it out with:

```
$ docker build -t ghcr.io/oasisprotocol/sapphire-localnet:local --build-arg VERSION=local -f docker/sapphire-localnet/Dockerfile .
$ docker run --rm -ti -p80:80 -p5432:5432 -p8544-8547:8544-8547 -v /tmp/eth-runtime-test:/serverdir/node ghcr.io/oasisprotocol/sapphire-localnet:local
```


docker/sapphire-localnet/Dockerfile
```diff
@@ -71,34 +71,34 @@ RUN wget https://github.com/oasisprotocol/${PARATIME_NAME}-paratime/releases/dow
...
-    && strip -S -x /nexus-git/nexus \
-    && mv /nexus-git/nexus / \
+ARG NEXUS_VERSION=0.4.1
+RUN wget https://github.com/oasisprotocol/nexus/releases/download/v${NEXUS_VERSION}/oasis_nexus_${NEXUS_VERSION}_linux_amd64.tar.gz \
+    && tar -zxvf oasis_nexus_${NEXUS_VERSION}_linux_amd64.tar.gz \
+    && strip -S -x /oasis_nexus_${NEXUS_VERSION}_linux_amd64/nexus \
+    && mv /oasis_nexus_${NEXUS_VERSION}_linux_amd64/nexus / \
+    && rm -rf /oasis_nexus_${NEXUS_VERSION}_linux_amd64* && \
+    # We also need to fetch the storage migrations.
+    git clone https://github.com/oasisprotocol/nexus.git nexus-git --branch v${NEXUS_VERSION} --depth 1 \
```
matevz, line 81: IMO this should be part of the nexus tar.gz (similar to how we ship oasis-node, oasis-net-runner, oasis-remote-signer etc. in the Oasis Core tarball).

matevz (approved):
While you're at it, can you also move the default explorer port from 80 to 8548. Since it's out of 8544-8547 range it requires passing another `-p` flag and port 80 is also often taken on the host. Moving the port would simplify the ports to `-p8544-85458:8544-8548`.

docker/sapphire-localnet/Dockerfile
ptrus: There are better solutions for this, for example embedding the migrations in the binary itself, so that separate files are not even needed. It's done like that here for the web3-gateway. Opened an issue in Nexus: https://github.com/oasisprotocol/nexus/issues/799

ptrus:
> While you're at it, can you also move the default explorer port from 80 to 8548

Sure, updated.

PR #653, opened 2024-11-21
feat(localnet): Add --no-explorer flag and reduce size

### LINKED ISSUE #640: sapphire-localnet: Add `--no-explorer` flag to disable Nexus and Explorer
Spinning up the web server and nexus indexer in the Localnet docker image causes additional delay and consumes CPU/memory/db cycles over longer run. For environments that don't need the indexer and the explorer (e.g. CI), add the `--no-explorer` flag to disable it.

TODO:
- [ ] implement `--no-explorer` flag
- [ ] benchmark startup times if they're reduced, see if you can speed it up more by shuffling start order
- [ ] also check, if there are any leftovers from indexer/explorer (are those stripped) so you can also reduce the image size
- [ ] besides updating Localnet docs in this repo, also update it here https://docs.oasis.io/dapp/tools/localnet
- [ ] I suggest we also change the explorer port from the standard `80` to `8548`. In practice you will always map the port to sth else, so the `Localnet Explorer available at http://localhost:80` message is incorrect. Also this reduces the number of `-p` docker parameters needed.

abukosek:
This PR:
- Adds the `--no-explorer` flag, which disables starting the explorer and indexer.
- Adds the `--no-gateway` flag, which disables starting the gateway.
- Reduces the size of the sapphire-localnet Docker image from 985MB to 932MB.
- Reduces the number of layers in the sapphire-localnet Docker image.
- Makes it easier to build components from unreleased branches (you no longer have to uncomment whole blocks of shell scripts, just a single line).
- Moves all the version variables to the top, so you don't have to dig through all the crap in the Dockerfile just to bump a version.
- Bumps CLI version to 0.10.3 and Nexus version to 0.4.2.

Closes #640.


docker/sapphire-localnet/Dockerfile
```diff
@@ -160,34 +177,36 @@ ENV OASIS_UNSAFE_MOCK_SGX=1
...
+    && rm -f /etc/apt/keyrings/envoy-keyring.gpg \
+    && rm -rf /root/.npm /root/.node-gyp \
+    && rm -rf /usr/share/doc
+
+# Copy all results from build step.
+COPY --from=oasis-core-dev /out /
+
+# Create symlinks for convenience and a dir to put all your rofls into,
+# also print version info.
+RUN ln -s ${OASIS_NODE_BINARY} ${OASIS_CLI_BINARY} /usr/local/bin \
```
matevz, line 215: Maybe also link the web3 gateway there for consistency?

matevz (commented):
First pass, I'll test it tomorrow.

docker/common/start.sh
```diff
@@ -18,6 +18,8 @@
 
 rm -f /CONTAINER_READY
 
+export OASIS_DOCKER_START_EXPLORER=${OASIS_DOCKER_START_EXPLORER:-yes}
+
 export OASIS_DOCKER_NO_GATEWAY=${OASIS_DOCKER_NO_GATEWAY:-no}
```
ptrus, line 23: Should we change this to `OASIS_DOCKER_START_GATEWAY`, for consistency with explorer?

docker/common/start.sh
```diff
@@ -148,6 +150,11 @@ while [[ $# -gt 0 ]]; do
       shift
       shift
       ;;
+    --no-explorer)
```
ptrus, line 153: Should we also add a flag for disabling gateway?

docker/common/start.sh
ptrus: Also I just remembered that enabling nexus/explorer while gateway is disabled wont work, so let's maybe catch that and error out (or alternative, if gateway is disabled also automatically disable nexus/explorer).

docker/common/start.sh
```diff
@@ -351,35 +358,37 @@ else
   sleep 10
 fi
 
-# Once everything is initialized and setup, start Nexus and Explorer.
-if [ ! -z "${OASIS_NEXUS_BINARY:-}" ]; then
-  notice "Creating database 'nexus'\n"
-  su -c "createdb -h 127.0.0.1 -p 5432 -U postgres nexus" postgres
+# Once everything is initialized and setup, start Nexus and Explorer if enabled.
+if [[ "${OASIS_DOCKER_START_EXPLORER}" == "yes" ]]; then
+  if [ ! -z "${OASIS_NEXUS_BINARY:-}" ]; then
```
ptrus: I think we can remove this check for the binary now and assume it's set if explorer is enabled? (like we do for gateway).

docker/sapphire-localnet/Dockerfile
```diff
@@ -1,144 +1,161 @@
...
+    && cd /oasis-core-km/tests/runtimes/simple-keymanager \
+    && OASIS_UNSAFE_ALLOW_DEBUG_ENCLAVES=1 CARGO_PROFILE_RELEASE_LTO=true CARGO_PROFILE_RELEASE_OPT_LEVEL=s CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1 OASIS_UNSAFE_SKIP_AVR_VERIFY=1 \
+    cargo build --release --features debug-mock-sgx \
+    && strip -s /oasis-core-km/target/release/simple-keymanager \
+    && mv /oasis-core-km/target/release/simple-keymanager /out \
+    && rm -rf /oasis-core-km \
+    \
+    && cd / \
+    && if [ "x${OASIS_CORE_BRANCH}" != "x" ]; then \
+        git clone https://github.com/oasisprotocol/oasis-core.git --branch ${OASIS_CORE_BRANCH} --depth 1 \
```
ptrus: Could we use a pattern that allows specifying either branch name or commit ref, and works in both cases?

I believe the clone->fetch->checkout as it is done for Explorer repository works with both. Maybe there's a more concise alternative, but it would be great if we would allow either branch name or specific commit refs to be specified for any component.

ptrus (commented):
A couple of minor suggestions. The updates to sapphire-localnet/Dockerfile are great, much clearer what versions of components are used.

docker/sapphire-localnet/Dockerfile
abukosek: I don't think that's useful, since the gateway is started at container startup if enabled and the user won't ever need to call it manually, whereas the CLI and node have useful commands.  This is probably why the original author of these links only linked the node and CLI :)

docker/common/start.sh
abukosek: Good catch, I'll fix that :)

docker/sapphire-localnet/Dockerfile
```diff
@@ -1,144 +1,161 @@
...
-#     && mv /nexus-git/nexus / \
-#     && mv /nexus-git/storage/migrations /nexus-migrations \
-#     && rm -rf /nexus-git
-
-# ARG EXPLORER_REPO=https://github.com/oasisprotocol/explorer.git
-# ARG EXPLORER_VERSION=v1.13.0
+#ARG NEXUS_BRANCH=main
+
 # TODO: Update to a released version once a version with localnet explorer support is available (https://github.com/oasisprotocol/explorer/issues/1597).
 ARG EXPLORER_REPO=https://github.com/oasisprotocol/explorer.git
```
ptrus: Feel free to remove the variable for the repository. I used it because my (very hacky) fork was used before explorer got some real support for localnet in https://github.com/oasisprotocol/explorer/issues/1597 , but then forgot to remove it 😁

abukosek: Oh, ok :D

docker/common/start.sh
matevz: I would simply disable both the explorer and indexer, if `--no-explorer` is passed.

.github/workflows/ci-benchmarks.yml
```diff
@@ -44,6 +44,7 @@ jobs:
       SAPPHIRE_PARATIME: ${{ github.workspace }}/oasis_core/sapphire-paratime
       OASIS_NODE_DATADIR: /tmp/oasis-sapphire-benchmarks
       OASIS_WEB3_GATEWAY: ${{ github.workspace }}/oasis-web3-gateway
+      OASIS_DOCKER_START_EXPLORER: no
```
matevz, line 47: Don't forget to describe the usage in the Sapphire Localnet chapter: https://github.com/oasisprotocol/docs/blob/main/docs/dapp/tools/localnet.mdx

matevz (commented):
While you're at it, can you also move the explorer port 80 to 8548 (see the last bullet https://github.com/oasisprotocol/oasis-web3-gateway/issues/640) and the docs accordingly.

docker/common/start.sh
abukosek: > I would simply disable both the explorer and indexer, if `--no-explorer` is passed.

That is already what happens, Peter suggested adding a similar option for the **gateway** :)

abukosek:
> While you're at it, can you also move the explorer port 80 to 8548 (see the last bullet #640) and the docs accordingly.

The port move seems to have already been done in https://github.com/oasisprotocol/oasis-web3-gateway/commit/a941da451734460cffdc6fc09d49dd382a65acde

.github/workflows/ci-benchmarks.yml
abukosek: PR opened in docs: https://github.com/oasisprotocol/docs/pull/1032

docker/common/start.sh
abukosek: > Should we change this to `OASIS_DOCKER_START_GATEWAY`, for consistency with explorer?

I've kept this as-is for compatibility, but we can do a separate PR to change this in the future.

abukosek:
Ready for re-review :)

In a future PR, we can replace nodejs with nginx and save about 200MB+.

docker/common/start.sh
```diff
@@ -351,8 +368,8 @@ else
   sleep 10
 fi
 
-# Once everything is initialized and setup, start Nexus and Explorer.
-if [ ! -z "${OASIS_NEXUS_BINARY:-}" ]; then
+# Once everything is initialized and setup, start Nexus and Explorer if enabled.
+if [[ "${OASIS_DOCKER_START_EXPLORER}" == "yes" ]]; then
```
ptrus, line 367: Can we also please update the notices at the end of initialization so that Web3 Gateway, Explorer and Nexus addresses are not displayed when they are not enabled? We currently always display the addresses.

Other than that, the rest looks good to me.

abukosek: Good catch!  Fixed :)

docker/common/start.sh
ptrus: Wait sorry, I was really stupid or something 😭 Nexus and Explorer will work just ok without web3-gateway? 

Can you please just revert this part where gateway automatically disables nexus/explrer, and then also at the end move 
```bash
  if [[ "${OASIS_DOCKER_START_EXPLORER}" == "yes" ]]; then
    notice "Nexus API listening on ${CYAN}http://localhost:8547${OFF}.\n"
    notice "Localnet Explorer available at ${CYAN}http://localhost:${EXPLORER_PORT}${OFF}.\n"
  fi
 ```
 
 out of the nested IF.

This case where you don't have gateway and do have nexus is probably not really useful since there likely wont be any Sapphire transactions, but one might still use it if it's interested only in the cosnseus part or something.

ptrus (approved):
Added one more comment, sorry I aperantly had some brain damage in the initial suggestion that web3 gateway should also disable nexus/explorer.

Otherwise looks good!

docker/common/start.sh
abukosek: No worries, fixed! :)

PR #657, opened 2024-11-26
feat: report just the number of logs matched in debug logs for `eth_getLogs`

ptrus:
Reduce verbosity of `eth_getLogs` debug logs: instead of logging full responses, only log the number of logs matched. Previously, we logged all logs in the queried range which really bloated the debug log output of the web3 gateway.

For debugging purposes, full responses can still be obtained by re-running the query, as the query itself is logged.


gw0 (approved):
Looks straightforward.

PR #659, opened 2024-12-02
sapphire: Keep unzip installed, needed for ROFLs

ptrus:
Unzip is needed to unpack orc archive, in case rofl's are used.


abukosek (approved):
Oops, my bad :)

ptrus:
Np, we should improve the CI tests, to catch these :)

PR #668, opened 2024-12-16
fix(localnet): don't transition to epoch 3

ptrus:
Transitioning to epoch should be enough.

There remains one (very) occasional issue (which I cannot reliably reproduce), where requests to keymanager fail with:
```
{"caller":"grpc.go:203","err":"rpc error: code = Unknown desc = key manager failure: call failed: key manager is not initialized","level":"error","method":"/oasis-core.RuntimeClient/Query","module":"grpc/internal","msg":"request failed","req_seq":5566,"ts":"2024-12-13T09:16:02.867419593Z"}
```

Update: The above issue is now resolved as well.

### DIFF
--- a/docker/common/start.sh
+++ b/docker/common/start.sh
@@ -315,16 +315,26 @@ if [[ "${BEACON_BACKEND}" == "mock" ]]; then
     # Ensure the key manager is fully ready before moving to the first epoch.
     notice_debug -l "Waiting for key manager to start up..."
     ${OASIS_NODE_BINARY} debug control wait-ready -a unix:${OASIS_KM_SOCKET}
+
+    # Store the keymanager's runtime extra_info to monitor for changes during epoch 1.
+    km_extra_info=$(${OASIS_NODE_BINARY} control status -a unix:${OASIS_KM_SOCKET} | jq -r '.registration.descriptor.runtimes[0].extra_info')
   fi
 
   echo -n .
   notice_debug -l "Setting epoch to 1..."
   ${OASIS_NODE_BINARY} debug control set-epoch --epoch 1 -a unix:${OASIS_NODE_SOCKET}
 
-  # Transition to the next epoch when the KM generates ephemeral secret.
   if [[ "${PARATIME_NAME}" == "sapphire" ]]; then
+    # Ensure ephemeral secret has been generated.
     notice_debug -l "Waiting for key manager to generate ephemeral secret..."
     while (${OASIS_NODE_BINARY} control status -a unix:${OASIS_KM_SOCKET} | jq -e ".keymanager.secrets.worker.ephemeral_secrets.last_generated_epoch!=2" >/dev/null); do
+      sleep 0.5
+    done
+
+    # Ensure the keymanager has re-registered with updated extra_info.
+    # This ensures that we wait until the master secret has been proposed and confirmed.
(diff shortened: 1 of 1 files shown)


docker/common/start.sh
```diff
@@ -315,16 +315,21 @@ if [[ "${BEACON_BACKEND}" == "mock" ]]; then
...
+    km_extra_info=$(${OASIS_NODE_BINARY} control status -a unix:${OASIS_KM_SOCKET} | jq -r '.registration.descriptor.runtimes[0].extra_info')
   fi
 
   echo -n .
   notice_debug -l "Setting epoch to 1..."
   ${OASIS_NODE_BINARY} debug control set-epoch --epoch 1 -a unix:${OASIS_NODE_SOCKET}
 
-  # Transition to the next epoch when the KM generates ephemeral secret.
+  # Transition to the next epoch only when the keymanager has re-registered with updated extra_info.
+  # This ensures that we wait until the master secret has been generated and published.
```
ptrus: Thanks to @peternose, we figured out the occasional remaining problem during the bootstrap. It was not sufficient to wait for ephemeral secret to be generated, but also that the master secrets are generated and published. We now do this by ensuring the KM re-registers with the updated extra_info.

docker/common/start.sh
```diff
@@ -315,16 +315,21 @@ if [[ "${BEACON_BACKEND}" == "mock" ]]; then
...
```
peternose:
```suggestion
  # This ensures that we wait until the master secret has been proposed and confirmed.
```

docker/common/start.sh
```diff
@@ -315,16 +315,21 @@ if [[ "${BEACON_BACKEND}" == "mock" ]]; then
...
   if [[ "${PARATIME_NAME}" == "sapphire" ]]; then
-    notice_debug -l "Waiting for key manager to generate ephemeral secret..."
-    while (${OASIS_NODE_BINARY} control status -a unix:${OASIS_KM_SOCKET} | jq -e ".keymanager.secrets.worker.ephemeral_secrets.last_generated_epoch!=2" >/dev/null); do
+    notice_debug -l "Waiting for key manager to generate secrets and re-register..."
+    # Ensure master secrets have been generated.
```
peternose:
```suggestion
    # Ensure master secret has been generated.
```

docker/common/start.sh
```diff
@@ -315,16 +315,21 @@ if [[ "${BEACON_BACKEND}" == "mock" ]]; then
...
 
   echo -n .
   notice_debug -l "Setting epoch to 1..."
   ${OASIS_NODE_BINARY} debug control set-epoch --epoch 1 -a unix:${OASIS_NODE_SOCKET}
 
-  # Transition to the next epoch when the KM generates ephemeral secret.
+  # Transition to the next epoch only when the keymanager has re-registered with updated extra_info.
+  # This ensures that we wait until the master secret has been generated and published.
   if [[ "${PARATIME_NAME}" == "sapphire" ]]; then
-    notice_debug -l "Waiting for key manager to generate ephemeral secret..."
```
peternose, line 326: You should still wait for ephemeral secret to be generated and proposed, as this is done concurrently. Although this will usually take less time than master secret generation and confirmation.

ptrus: Thanks, updated

peternose (approved):

PR #676, opened 2024-12-24
other(localnet): replace nodejs with nginx

### LINKED ISSUE #658: Replace nodejs with nginx in localnet explorer
In a future PR, we can replace nodejs with nginx and save about 200MB+.

_Originally posted by @abukosek in https://github.com/oasisprotocol/oasis-web3-gateway/issues/653#issuecomment-2498007081_

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-web3-gateway/issues/658

Reduces sapphire image size from ~970MB to ~793MB.

### DIFF
--- a/docker/common/start.sh
+++ b/docker/common/start.sh
@@ -15,6 +15,7 @@
 # - OASIS_NEXUS_BINARY: path to oasis-nexus binary
 # - OASIS_NEXUS_CONFIG_FILE: path to oasis-nexus config file. Required if OASIS_NEXUS_BINARY is provided
 # - OASIS_EXPLORER_DIR: path to explorer (nexus frontend) directory
+# - OASIS_EXPLORER_NGINX_CONFIG_FILE: path to explorer nginx config file.
 
 rm -f /CONTAINER_READY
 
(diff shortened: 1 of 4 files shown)


abukosek (approved):
Awesome!

matevz:
Could we also replace envoy with nginx?

docker/common/start.sh
```diff
@@ -15,6 +15,7 @@
 # - OASIS_NEXUS_BINARY: path to oasis-nexus binary
 # - OASIS_NEXUS_CONFIG_FILE: path to oasis-nexus config file. Required if OASIS_NEXUS_BINARY is provided
 # - OASIS_EXPLORER_DIR: path to explorer (nexus frontend) directory
+# - OASIS_EXPLORER_NGINX: path to explorer nginx config file.
```
matevz: I would prefer the longer naming for consistency.

```suggestion
# - OASIS_EXPLORER_NGINX_CONFIG_FILE: path to explorer nginx config file.
```

ptrus:
> Could we also replace envoy with nginx?

Not if we want to support grpc-web

kostko (approved):

docker/common/start.sh
ptrus: 👍  Updated

PR #742, opened 2025-06-05
other: use slices.Contains to simplify code

rustrover:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

### DIFF
--- a/filters/event_system.go
+++ b/filters/event_system.go
@@ -3,6 +3,7 @@ package filters
 import (
 	"fmt"
 	"math/big"
+	"slices"
 	"sync"
 	"time"
 
@@ -194,16 +195,6 @@ func (es *EventSystem) SubscribeNewHeads(headers chan *types.Header) *Subscripti
 	return es.subscribe(sub)
 }
 
-func includes(addresses []common.Address, a common.Address) bool {
-	for _, addr := range addresses {
-		if addr == a {
-			return true
-		}
-	}
-
-	return false
-}
-
 // filterLogs creates a slice of logs matching the given criteria.
 func filterLogs(logs []*types.Log, fromBlock, toBlock *big.Int, addresses []common.Address, topics [][]common.Hash) []*types.Log {
 	var ret []*types.Log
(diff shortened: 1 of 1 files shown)


filters/event_system.go
```diff
@@ -195,13 +196,7 @@ func (es *EventSystem) SubscribeNewHeads(headers chan *types.Header) *Subscripti
...
 
 func includes(addresses []common.Address, a common.Address) bool {
-	for _, addr := range addresses {
-		if addr == a {
-			return true
-		}
-	}
-
-	return false
+	return slices.Contains(addresses, a)
```
ptrus: Can you please remove `includes` method, and just inline the `slices.Contains` call instead.

filters/event_system.go
```diff
@@ -224,13 +219,7 @@ Logs:
...
 		}
 		for i, sub := range topics {
-			match := len(sub) == 0 // empty rule set == wildcard
-			for _, topic := range sub {
-				if log.Topics[i] == topic {
-					match = true
-					break
-				}
-			}
+			match := slices.Contains(sub, log.Topics[i])
```
ptrus: if `sub` is empty, `match` should also be `true`. So i think this should be:

```go
if  len(sub) == 0 || slices.Contains(sub, log.Topics[i]) {
  continue Logs
}
```

ptrus (changes requested):

filters/event_system.go
rustrover: Thanks. Done.

filters/event_system.go
rustrover: Modified. Thanks for your review.

rustrover: Thank you for your review. However, it seems that the logic here is actually the opposite.

ptrus:
Looks good thanks! Please rebase on master and its good to merge!

ptrus (approved):

rustrover:
> Looks good thanks! Please rebase on master and its good to merge!

Done. Please review again.

ptrus:
Thanks!

PR #795, opened 2025-10-09
Add advance-epoch.sh script to localnet Docker images

abukosek:
This PR adds a helper script to the localnet Docker images, `/advance-epoch.sh`, which advances the mock epoch by 1 or by the argument given.

### DIFF
--- a/docker/common/advance-epoch.sh
+++ b/docker/common/advance-epoch.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+#
+# This script advances the mock epoch by 1 or by the number of epochs
+# given as the first argument to this script.
+
+set -eo pipefail
+
+num_epochs="${1:-1}"
+
+OASIS_NODE_SOCKET=${OASIS_NODE_DATADIR}/net-runner/network/client-0/internal.sock
+
+if [[ "${BEACON_BACKEND:-mock}" != "mock" ]]; then
+  echo "This command only works with the mock beacon backend."
+  echo "Run the container with BEACON_BACKEND set to 'mock' to enable it."
+  exit 1
+fi
+
+# Wait for the Oasis Node to start.
+while [[ ! -S ${OASIS_NODE_SOCKET} ]]; do sleep 1; done
+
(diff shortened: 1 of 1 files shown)


docker/common/advance-epoch.sh
```diff
@@ -0,0 +1,20 @@
...
+set -euo pipefail
+
+num_epochs="${1:-1}"
+
+OASIS_NODE_SOCKET=${OASIS_NODE_DATADIR}/net-runner/network/client-0/internal.sock
+
+# Wait for the Oasis Node to start.
+while [[ ! -S ${OASIS_NODE_SOCKET} ]]; do sleep 1; done
+
+old_epoch=`${OASIS_NODE_BINARY} control status -a unix:${OASIS_NODE_SOCKET} | jq '.consensus.latest_epoch'`
```
ptrus, line 21: Maybe ensure that mock epoch backend is configured (I think this can be seen via control status iirc), and if mock backend is not used print a nice error to the user.

ptrus (commented):
👍  just one suggesrion

docker/common/advance-epoch.sh
abukosek: I've checked and unfortunately that isn't exported in the control status :(
I can check the `BEACON_BACKEND` variable though -- if it's set and if it's set to anything other than `mock`, the error message will be printed, otherwise the start script assumes a mock backend.

ptrus: oh for some reason i thought we expose the consensus parameters. Maybe could check via `oasis-cli network show`?  

But your suggestion is fine with me as well 👍

ptrus (approved):

docker/common/advance-epoch.sh
abukosek: In the control status output there's just the consensus backend (which is always `tendermint`) :man_shrugging:

The check for the `BEACON_BACKEND` variable that I just added should work fine, since the startup script also uses that :)
(And it's much simpler than parsing output from the CLI.)

PR #797, opened 2025-10-24
docker: Bump Oasis Core to 25.5 and CLI to 0.16.0 for sapphire-localnet

matevz:
This matches current Mainnet parameters.
Also changes postgres to docker.io/postgres so the images are buildable under podman.

### DIFF
--- a/docker/sapphire-localnet/Dockerfile
+++ b/docker/sapphire-localnet/Dockerfile
@@ -1,7 +1,7 @@
 ARG PARATIME_VERSION=1.0.0-testnet
 ARG PARATIME_NAME=sapphire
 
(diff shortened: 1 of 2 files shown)


docker/sapphire-localnet/Dockerfile
```diff
@@ -1,7 +1,7 @@
 ARG PARATIME_VERSION=1.0.0-testnet
```
ptrus, line 1: Maybe also update to `ARG PARATIME_VERSION=1.0.0`

ptrus (approved):

ptrus:
Updated in https://github.com/oasisprotocol/oasis-web3-gateway/pull/799 (also some other versions)

